Issue #162, opened 2020-02-20
add support for Syriac


balmas:
Morphology and glosses provided by the Sedra API 

http://sedra.bethmardutho.org/about/openapi

balmas:
Still TODO before we can release this:

- [ ] fix the formatting of the sources  
- [ ] verify the content in Credits with Bethmardutho project team
- [ ] formatting of syriac chars in the short definitions
- [ ] decide whether or not to include all the short definitions by default (multilingual defs are available)
- [ ] decide whether or not it needs to be flagged as experimental

balmas:
for sources, George would like to try something like the following (modeled after the ge'ez approach)

1. SyriacWord1 [SedraLink]
       part-of-speech
       gloss1, gloss2, gloss3 (Dictionary1Abbreviation); gloss4, gloss5, gloss6 (Dictionary2Abbreviation)
2. ditto

Credits
 Sedra [with a link to sedra.bethmardutho.org], Georg A. Kiraz (General Editor), James W. Bennet (IT Manager)
 Dictionary1Abbreiviaiton = <full citation>
 Dictionary2Abbreivation = <full citation>
(but don't duplicate if an abbreviation appears more than once and list them in alphabetical order of the abbreviations)

balmas:
Note, this will be released as an experimental language. Credits fixes require changes on the Sedra side and will be done in a future release.

balmas:
Alpheios Components 3.3.0-qa.20200513442

@monzug a good place to look for Syriac words that *should* produce results in Alpheios are the texts at http://sedra.bethmardutho.org/library

monzug:
Credits: I don't see the link to sedra site (as said above it will be done in the future)

I have a question on the repetition of the source for a word that is both a verb and a noun. the verb is repeated few times with same source. can it be grouped together? ܫܩܠ     word

![syriac-source](https://user-images.githubusercontent.com/41396793/82330625-16dc0300-99b1-11ea-80c3-058758893e7f.JPG)

balmas:
> I have a question on the repetition of the source for a word that is both a verb and a noun. the verb is repeated few times with same source. can it be grouped together? ܫܩܠ word
> 

This is an issue with the Sedra API - let's track it via a new issue #355

PR #170, opened 2020-03-03
wip on #162

balmas:
allow html in sedra short defs


balmas:
> Looks well!
> I could see problems with tests for UserDataManager and that tests are much less then before.
> I believe that it is ok for now.

hmm. i don't think those test failures could be related to this change, so I'll merge it but we'll have to investigate what's happening with the tests.

balmas:
> > Looks well!
> > I could see problems with tests for UserDataManager and that tests are much less then before.
> > I believe that it is ok for now.
> 
> hmm. i don't think those test failures could be related to this change, so I'll merge it but we'll have to investigate what's happening with the tests.

the tests do pass locally, so maybe a travis environment problem?

PR #175, opened 2020-03-04
Add id to some Vue components for selenium tests

irina060981:
For selenium tests


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "alpheios-core",
-  "version": "1.0.2",
+  "version": "1.0.2-dev",
```
balmas: I think we have dropped this approach to versioning now. It should just be 1.0.2

irina060981: I changed it manually, I thought that it would be right.
Will change

Issue #165, opened 2020-02-24
Persian lookup bug


irina060981:
>lookup (change language to Persian from drop-down and enter جان    check that the word is entered from right to left in the lookup field and that in the morphology popup its meaning is father or soul. close popup)

There is a problem with Persian language - a lookup (from lookup input) produces result only after the second or third lookup (in any browser).
It is a well-known problem, but I am not sure that I should define to lookup three times to get the result.

_Originally posted by @irina060981 in https://github.com/alpheios-project/selenium-tests/issues/5#issuecomment-589574479_

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200309431

monzug:
tested few times on chrome. when I lookup for جان , it takes sometimes (few seconds) before generating the definitions. in the meantime it does still display the message morphological data not found. so it does work somehow but t would be great if we can speed up the display of the definitions. is it possible?

balmas:
@monzug it is super fast for me after the first lookup, which makes sense to me because with the first lookup it needs to load the entire index file, but after that it should be a quick lookup in memory for it and any other persian word. I'm puzzled why it would be slow for you. I am not thrilled with the morphological data message displaying but can live with it for now.

monzug:
it's the first lookup that is super slow. any subsequential lookup is fast. if this is ok, I can close this one

monzug:
see screenshots of the popups on first lookup for جان 
![persian-2](https://user-images.githubusercontent.com/41396793/76526462-ee310d80-6443-11ea-8cac-557072bcced7.JPG)
![IMG_4372](https://user-images.githubusercontent.com/41396793/76526545-13be1700-6444-11ea-847f-221f20af12a0.jpg)

balmas:
it seems like there are still some timing issues with the notifications.  We might also need to increase the timeout on the initial request. Giving back to @irina060981 to see what she thinks.

irina060981:
Fixed, merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200320403

monzug:
I don't get anymore the message جان not found in our Persian resources in the morphology pop-up. However I do still experience the message reported in the first screenshot above.
anything to improve this?

balmas:
@monzug do you mean the Lexical Query produced no results message?

monzug:
[...]

Issue #157, opened 2020-02-10
css stylesheet  needs to be updated in v3.0.6-qa.3 build


monzug:
Just noticed that in build  Alpheios Reading Tools 3.0.6.99
Alpheios Components 1.6.3-dev  we are again missing the space in the morphology lookup
e.g. personsing
see attachment
![personsing](https://user-images.githubusercontent.com/41396793/74158820-bdfa1300-4be8-11ea-8d0f-0ce7280e8061.JPG)

monzug:
we did also lost the style for the verbs.  see attachment

![Screen Shot 2020-02-16 at 6 26 30 PM](https://user-images.githubusercontent.com/41396793/74609400-fd919500-50e9-11ea-8934-f29cd41263e7.png)

balmas:
I think this is fallout from https://github.com/alpheios-project/components/pull/894

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200309431

monzug:
both issues reported upthere are still reproducible in build  Alpheios Reading Tools 3.1.0 build qa.20200309459
Alpheios Components 3.3.0-qa.20200309431 build qa.20200309431 
tested in FF

balmas:
@irina060981 can you take another look at this? Thanks!

monzug:
one more in Chrome
![short-def](https://user-images.githubusercontent.com/41396793/76338016-87d6ae80-62ce-11ea-94b9-3bdd42266ecb.JPG)

irina060981:
>we did also lost the style for the verbs. see attachment

I didn't see the problems with styles in inflections, but according to the old version in https://texts.alpheios.net/
inflection table for latin irregular verbs are broken

![image](https://user-images.githubusercontent.com/12377640/76590875-4b0eef80-653a-11ea-9dfb-a60de75b0d9a.png)

@monzug , do you mean this problem?

monzug:
Irina, here is a screenshot in build Alpheios Reading Tools 3.1.0 build qa.20200309459
Alpheios Components 3.3.0-qa.20200309431 build qa.20200309431
![verb-latin-irreg](https://user-images.githubusercontent.com/41396793/76596697-38ae9a80-64d5-11ea-9406-3353408991bd.JPG)

irina060981:
Monica, I am sorry - but I don't see here styles problems.
[...]

PR #180, opened 2020-03-05
Fix bugs with persian lookup

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/165

There were several problems solved:

1. As Persian language has two iterations (first for definitions and only then to morph - canReset = true), it executed two parallel lexicons request (one request for each itertion) - this way it uploaded definitions from source two times

2. Persian words (and any other that could be filled with definitions later) had messed notifications, sometimes is not found, sometimes Definitions query pending

3. Popup with persian words was not updated after definitions were recieved


packages/components/dist/alpheios-components.js
```diff
@@ -66906,41 +66915,47 @@ class LexicalQuery extends _query_js__WEBPACK_IMPORTED_MODULE_1__["default"] {
...
 
-    const adapterLexiconResShort = yield alpheios_client_adapters__WEBPACK_IMPORTED_MODULE_3__["ClientAdapters"].lexicon.alpheios({
-      method: 'fetchShortDefs',
-      clientId: this.clientId,
-      params: {
-        opts: lexiconShortOpts,
-        homonym: this.homonym,
-        callBackEvtSuccess: LexicalQuery.evt.SHORT_DEFS_READY,
-        callBackEvtFailed: LexicalQuery.evt.SHORT_DEFS_NOT_FOUND
+    if (!this.startedDefinitionsQueries) {
```
balmas: the problem with this is that we sometimes DO want to run the definitions queries a 2nd time. We do this crazy 2nd iteration with the persian lookup because the persian morphological parser is a little limited and the dictionary does support some full forms as well as lemmas. So, we first lookup to see if there is an entry for the full form of the persian word, and if we don't get results, then we try again using the morphological parser to parse the lemma and then we try looking up the lemma. Often it will be the same as the full form, but occasionally the parser adds some value here and the 2nd lookup in the dictionary is for a different word than the first and we get a result.

Eventually I think we need to move all of this business logic behind looking up words to a separate service, probably behind a GraphQL api so that we can control and fine tune it more easily. 

In the meantime, maybe checking to see if we started queries *for the same homonym already* is what we need to do.

irina060981: I didn't think that we have such a specific workflow for Persian inside LexicalQuery workflow.
Ok - will add check for targetWord.
The main problem here - that getting data from Persian is too long - and this leads to multiple upload from Persian. May be it is worth to add check for multiple request to client-adapters.

irina060981:
I have added check for multiple upload resources to client-adapters and add aditional check to Lexical Query.

irina060981:
I added a very small fix for the issue 
https://github.com/alpheios-project/alpheios-core/issues/181

irina060981:
Added a small fix for https://github.com/alpheios-project/alpheios-core/issues/157
image

PR #185, opened 2020-03-09
fix alpheios-project/embed-lib#106

balmas:
app-authenticator and session-authenticator moved to components


balmas:
> Looks good to me (did not check the code of the authenticatior class assuming it stayed the same).

yep, all that changed was the import path for AuthData

Issue #177, opened 2020-03-04
stub the lexicon client-adapter tests


balmas:
the client-adapter lexicon tests are timing out sometimes in travis.

we should probably not be loading the full index files from the remote server for these tests. we need to stub them and add fixtures for that data similar to what we have done for other adapters.

for now the client-adapter tests have been removed from the test-ci script.  We need to put them back once this is fixed.

irina060981:
Merged and closed

PR #191, opened 2020-03-11
Use fixtures in lexicons tests

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/177

Added fixtures the following way:

All fixtures are added to inner cache and then it is used for short, index and checked for full definitions.
I think that it is the simplest and the most light way to include fixtures to lexicons.

After fixtures PR - https://github.com/alpheios-project/fixtures/pull/25
would be merged and uploaded we could turn on client-adapters tests.


packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -147,13 +147,21 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
   * @param {String} url - url from what we need to cache data
   * @return {Boolean} - true - if cached is successed
   */
-  async checkCachedData (url) {
+  async checkCachedData (url, externalData = null, skipFetch = false) {
```
balmas: please add the new parameters to the JSDOC

packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -250,7 +257,12 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
   */
   async updateFullDefs (fullDefsRequests, config, homonym) {
     for (let request of fullDefsRequests) { // eslint-disable-line prefer-const
-      const fullDefDataRes = this.fetch(request.url, { type: 'xml' })
+      let fullDefDataRes
+      if (cachedDefinitions.has(request.url)) {
```
balmas: so were we never checking the cached definitions before?

irina060981: Yes, there were not such caching for full definitions before.
It would be useful both for definitions and for fixtures.

packages/client-adapters/src/adapters/lexicons/adapter.js
irina060981: Missed that

irina060981: Fixed

irina060981:
ok - I would merge it for now, and will update after fixtures would be merged

Issue #176, opened 2020-03-04
reenable the wordlist user data manager tests


balmas:
timeouts (or something) are causing them to fail on travis. they are commented out to get the build/ci process working but we need to figure out how to get these going. We probably need to use stubs for the remote service.

irina060981:
Merged and closed

PR #197, opened 2020-03-17
Add fixtures to UserDataManager

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/176

1. Update lexicons fixtures to be used out of client-adapters repo
2. Use lexicons fixtures inside wordlist repo
3. Use fixtures for morphology inside concordance adapter tests

I think that if it would be one approval - it would be enough, but point both - because only one of you would have time for that :)


irina060981:
ok - than will merge it :)

PR #199, opened 2020-03-17
Lazy loading of CEDICT data

kirlat:
Updated components to load CEDICT data lazily. 

If LexicalQuery sends a `getWords` request and CEDICT data is not loaded yet, a response from a CEDICT service will arrive with a `SERVICE_UNINITIALIZED` error code. A LexicalQuery will recognize this error code and will notify a Lexis data module. In response to that, a Lexis data module will issue a `loadData` request to a CEDICT service. Once the successful response to a `loadData` request is received by a Lexis data module, we know that CEDICT data is available and we can notify user to repeat his/her lexical request.


packages/client-adapters/src/errors/cedict-error.js
```diff
@@ -0,0 +1,16 @@
+export default class CedictError extends Error {
```
balmas: what do you think about naming this something more generic, like CodedError (or something bettter if you can come up with it)? It's not really specific to Cedict is it? It's just a type of a error that has a code associated with it. We could then reuse the same class in other circumstances.

packages/client-adapters/src/adapters/base-adapter.js
```diff
@@ -27,6 +28,11 @@ class BaseAdapter {
     this.errors.push(error)
   }
 
+  addCedictError (errorCode, message) {
```
balmas: see my comment below .. I think this should be something generic and not specific to Cedict

packages/components/src/styles/skins/_alpheios.scss
```diff
@@ -283,8 +283,11 @@ html {
   --alpheios-notification-important-bg: rgba(198, 73, 6, 0.31);
   --alpheios-notification-important-border-color: var(--alpheios-important-notification-color);
 
-  //Hint
+  --alpheios-notification-imporant-alt-color: #31595E;
+  --alpheios-notification-imporant-alt-bg: #B0C38D;
+  --alpheios-notification-imporant-alt-border-color: #31595E;
```
balmas: just a typo in the style name - should be 'important' not 'imporant'

packages/components/tests/helpclasses/base-test-help.js
```diff
@@ -17,414 +18,416 @@ import AuthModule from '@/vue/vuex-modules/data/auth-module.js'
...
+        [enUSInfl, Locales.en_US],
+        [enGB, Locales.en_GB]
+      ])
+    })
+  }
+
+  static lexisModule (store, api) {
+    return new LexisModule(store, api, {
+      name: 'cedict',
+      targetURL: 'https://lexis-dev.alpheios.net/index-dev.html',
```
balmas: does this create a test-time dependency on the service running there?

packages/components/src/lib/auth/app-authenticator.js
```diff
@@ -105,12 +105,14 @@ export default class AppAuthenticator {
           // Unrecoverable error handler
           this.auth0Lock.on('unrecoverable_error', (error) => {
             console.error('Auth0 Lock unrecoverable error: ', error)
+            // eslint-disable-next-line prefer-promise-reject-errors
```
irina060981: Do you really have this error (es-lint promise-reject-error) ?
I have fixed it in the previous PR with adding new Error.
Why did you add this exception?

packages/components/src/vue/components/notification-area.vue
```diff
@@ -41,6 +41,32 @@
           <close-icon/>
       </div>
     </div>
+    <div
```
irina060981: Kirill, what do you think about splitting notification area content to several components.
For now it has such a specific structure, that we should create a specific separate area for each notification:
hint, notification, auth notification and now cedict notification.
This component becomes bigger and bigger, it has duplicted code and it is not easy to ready.
May be it is a good time to optimize it?
Or I could do it while would work on notification component for toolbar.
@kirlat , @balmas , what do you think?

packages/client-adapters/src/errors/cedict-error.js
kirlat: I think that's a great idea, I will think about what to rename it to.

packages/client-adapters/src/adapters/base-adapter.js
kirlat: Agree

packages/components/src/styles/skins/_alpheios.scss
kirlat: Oh, that's a good catch, will rename

packages/components/tests/helpclasses/base-test-help.js
kirlat: This is an important consideration and I agree with your concerns. It is not used in tests now (i.e. we're not testing a remote connectivity of Lexis CS at the moment) so I removed those config props. I'm not sure what would be the best way, but maybe we should use fixtures for things like this in the future.

packages/components/src/lib/auth/app-authenticator.js
kirlat: Oh, probably you've fixed it upstream, but I was working with the code based on the earlier version of master. So I've seen the error, suppressed it, and only after that merged your changes with the fixes. So I will remove those as they are no longer needed. Thanks for catching this!

packages/components/src/vue/components/notification-area.vue
kirlat: I think it would be great to split notifications into several components, so each single notification would be represented by a single instance of that component. With more and more parts of an application generating notifications we should have an architecture where each part (such as CEDICT) would be able to create a notification easily.

I think it's also good for notifications to be standardized using the same overall design. However, notifications may differ in colors. Each notification might have several elements that could be turned on or off for each particular notification instance. Those elements might include a text message, a progress bar, one or several action buttons, a close icon, and whatever else we might find worth including.

What do you think?

packages/client-adapters/src/errors/cedict-error.js
kirlat: @balmas, how about `RemoteError`? Errors in client adapters are classified by what type of error it is, not what format it has. Following this logic, a `RemoteError` will be an error generated by a remote server. Since a `ResponseMessage` now includes an error code if something goes wrong, we can say that a `RemoteError` will always have an error code, even when other services (i.e. non-CEDICT) will be used. 

Does it make sense?

balmas: Yes I like this idea

packages/components/src/vue/components/notification-area.vue
balmas: yes I agree with these suggestions. Irina if you want to start working on the notification component refactoring this is fine with me.

packages/components/tests/helpclasses/base-test-help.js
balmas: probably yes

Issue #195, opened 2020-03-13
irregular verb table broken


balmas:
>we did also lost the style for the verbs. see attachment

I didn't see the problems with styles in inflections, but according to the old version in https://texts.alpheios.net/
inflection table for latin irregular verbs are broken

![image](https://user-images.githubusercontent.com/12377640/76590875-4b0eef80-653a-11ea-9dfb-a60de75b0d9a.png)

@monzug , do you mean this problem?

_Originally posted by @irina060981 in https://github.com/alpheios-project/alpheios-core/issues/157#issuecomment-598547702_

balmas:
Alpheios Components 3.3.0-qa.20200320403

monzug:
Now the irregular tables are displayed correctly. tested in build qa.20200320411

Issue #75, opened 2020-01-15
inflection browser shows indefinite data for interrogative pronoun


balmas:
I think the standard form data for the greek interrogative pronoun table is incorrect - it's showing the indefinite table instead of the interrogative

![Screenshot from 2020-01-15 15-07-31](https://user-images.githubusercontent.com/644870/72467369-c6595c80-37a8-11ea-9039-27f8fb36dcb3.png)

balmas:
Alpheios Components 3.3.0-qa.20200320403

monzug:
much better, still I believe there are some clean up needed in the Browse Indefinite Pronoun Decl.
see attachment
![Screen Shot 2020-03-23 at 4 11 28 PM](https://user-images.githubusercontent.com/41396793/77332238-18c37600-6d22-11ea-8d2d-4746ae9435d2.png)

balmas:
ok thanks.

balmas:
Alpheios Components 3.3.0-qa.20200327479

monzug:
looks good in Alpheios Reading Tools 3.1.0 build qa.20200330265
Alpheios Components 3.3.0-qa.20200327479

Issue #64, opened 2019-12-10
Greek Interrogative pronouns - no pronoun views


irina060981:
While working on https://github.com/alpheios-project/inflection-tables/issues/289
I was not able to find examples for getting Interrogative pronouns.

For **τι**, 
Indefinite Pronoun Declension

For **ποιος**, **ποια**, **ποιο**
No lexical result

For **τίνος**, **τίνων**
returns irregular instead of pronoun

![image](https://user-images.githubusercontent.com/12377640/70515342-eb8ef580-1b80-11ea-9267-364fbcae6a7f.png)

![image](https://user-images.githubusercontent.com/12377640/70515411-05303d00-1b81-11ea-9180-ea8c14511f96.png)

For **πόσος**, **πόση**
returns adjective

![image](https://user-images.githubusercontent.com/12377640/70515495-255ffc00-1b81-11ea-8a34-2f268e3c512d.png)

![image](https://user-images.githubusercontent.com/12377640/70515571-49bbd880-1b81-11ea-8f9b-50bad56db3da.png)

For  **πόσο**
returns no lexical results

So I am not sure if Greek Interrogative pronouns works.
@balmas, I have created issue here not to forget about this problem.

balmas:
Alpheios Components 3.3.0-qa.20200320403

monzug:
Now it works fine for some pronouns.
there is still some clean up to improve the view of the pronoun:
1) it looks like the relative and indefinite tables in the browser are identical, which I did mention in GitHub 75
2) if I enter ἀττα from Vanessa's indefinite pronoun list, it does display the Greek Pronoun declension which we do not have in the browse inflection tables, then the Reflexive, Demonstrative, Personal pronoun tables. so, why do we still have the Greek Pronoun table?
this is the output from Morpheus:
<hdwd xml:lang="grc">τις</hdwd>
<pofs order="5">pronoun</pofs>
</dict>
<infl>
<term xml:lang="grc">
<stem>ἄσσα</stem>
</term>
<pofs order="5">pronoun</pofs>
<case order="4">accusative</case>
<gend>neuter</gend>
<num>plural</num>
<dial>epic</dial>
<stemtype>indef</stemtype>
<morph>enclitic indeclform</morph>
</infl>

see attachment
![Screen Shot 2020-03-23 at 4 42 11 PM](https://user-images.githubusercontent.com/41396793/77335880-0d267e00-6d27-11ea-89ff-e53101343d8a.png)

[...]

Issue #37, opened 2018-09-25
make greek pronoun matching more precise


balmas:
we currently use the pronoun table data to identify the class of a pronoun. This can lead to incorrect results if a particular form appears in multiple tables.

per alpheios-project/inflection-tables#166 `οἷ` appears in both the personal pronoun and relative pronoun tables but it's not clear if this is correct.

Also, this produces morphology matches on incorrect cells in each of the personal and relative tables because the parser results have fewer features than the table data.

balmas:
not closed. caught in the cross-fire of repository merging.

PR #202, opened 2020-03-19
I75

balmas:
This fixes the following issues

#195
#75 
#64 
and at least part of #37 (showing only the correct tables for οἷ)


packages/components/package.json
```diff
@@ -45,7 +45,7 @@
     "alpheios-client-adapters": "^3.3.0",
     "alpheios-data-models": "^3.3.0",
     "alpheios-experience": "github:alpheios-project/experience",
-    "alpheios-fixtures": "github:alpheios-project/fixtures",
+    "alpheios-fixtures": "github:alpheios-project/fixtures#i75-add-ti/s",
```
irina060981: I believe, it would be removed

packages/inflection-tables/tests/data/greek-views.test.js
```diff
@@ -7,18 +7,18 @@ import { Constants, Feature, LanguageModelFactory } from 'alpheios-data-models'
 import BaseTestHelp from '@tests/data/base-test-help.js'
 
 describe('greek-views.test.js', () => {
-  console.error = function () {}
-  console.log = function () {}
-  console.warn = function () {}
+  //console.error = function () {}
```
irina060981: I think you could remove this comments signs

packages/client-adapters/src/adapters/tufts/engine/sedra.js
```diff
@@ -11,7 +11,7 @@ data.setMeaningParser(function (meaning, targetWord) {
   return new Models.Definition(meaningText, lang, 'text/plain', targetWord)
 })
 
-data.setPropertyParser(function (propertyName, propertyValue) {
+data.setPropertyParser(function (propertyName, propertyValue, inputElem) {
```
irina060981: Bridget, I am sorry, may be I did miss something.
I see how you added inputElem to arguments of parserProperty - but I don't see any usage inside functions of this argument.
May be it is not needed?

irina060981:
Also you have some errors for inflections in latin set inside tests.

packages/components/package.json
balmas: oh, shoot I thought I caught all of these. Thanks, I will fix it!

packages/inflection-tables/tests/data/greek-views.test.js
balmas: good catch, thanks! will fix.

packages/client-adapters/src/adapters/tufts/engine/sedra.js
balmas: I changed the function signature because I needed the input element in the morpheusgrc engine to check if the property was in the hdwd

irina060981:
Bridget, I have uncomitted user-data-manager tests - because I couldn't reproduce fail of them on my own branch.
if they will fail again - and you will need to comment them again - feel free to do that - I will check it again later

balmas:
> Also you have some errors for inflections in latin set inside tests.

oh that's weird. i thought they were all passing. Maybe something changed after I merged master in. I will double check. Thanks!

PR #207, opened 2020-03-24
I75 fix

### LINKED ISSUE #75: inflection browser shows indefinite data for interrogative pronoun

balmas:
fixing forms used for standard form tables for inflection browser
Greek relative and indefinite pronouns

fixes #75 and part of #64


balmas:
data change only so will merge without review.

PR #212, opened 2020-03-27
Small fix for span padding

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/157
image

image


irina060981:
@balmas - it is a very tiny fix - I will merge it without an approval

Issue #215, opened 2020-03-27
Fix tests for inflection browser


kirlat:
It seems that tests for `inflection-browser.vue` in `components` are currently not working after refactoring. In #214 I've added them to exclusions but I think we eventually need to fix them.

irina060981:
oh, I have missed that - will fix

irina060981:
Merged and closed

PR #218, opened 2020-03-30
Fix inflection browser tests

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/215

I have placed tests to separated threads for travis - this way user-data-manager tests start to work well, and I was able to add tests for inflection browser

I was to switch off res-client tests - because for some resons jest failed down after all 4 tests were completely done. I believe that this repo would be anyway refactored - so I switched them off for now.

@kirlat , I am not well familiar with travis deploy tasks - so could you check if our previously used format doesn't break deploy task.
@balmas , I suggest to use tests way

@kirlat and @balmas , would be glad to get your feedback :)


package.json
```diff
@@ -15,17 +15,17 @@
...
+    "test-wordlist": "npm --prefix packages/wordlist run test-no-coverage",
     "build-client-adapters": "npm --prefix packages/client-adapters run build",
-    "test-client-adapters": "npm --prefix packages/client-adapters run test",
+    "test-client-adapters": "npm --prefix packages/client-adapters run test-no-coverage",
     "build-inflection-tables": "npm --prefix packages/inflection-tables run build",
-    "test-inflection-tables": "npm --prefix packages/inflection-tables run test",
+    "test-inflection-tables": "npm --prefix packages/inflection-tables run test-no-coverage",
     "build-components": "npm --prefix packages/components run build",
-    "test-components": "npm --prefix packages/components run test-ci",
+    "test-components": "npm --prefix packages/components run test-no-coverage",
```
kirlat: I'm wondering why do we run tests without coverage reporting? Coverage reporting may not work in local environment (it does not work for me at least) but I think this can work in Travis environment and is useful there.

packages/components/tests/inflections/inflections-browser-1.test.js
```diff
@@ -76,7 +77,8 @@ describe('inflections.test.js', () => {
     expect(titles.at(8).text()).toEqual(expect.stringContaining('Verb Paradigms'))
 
     let wideTables = cmp.findAll(InflectionsTableWide)
-    expect(wideTables.length).toEqual(118)
+    // expect(wideTables.length).toEqual(118)
+    console.info('wideTables - ', wideTables.at(0).props().view.constructor.name)
```
kirlat: Do we need this info statement for testing?

.travis.yml
```diff
@@ -35,3 +35,11 @@ deploy:
...
       all_branches: true
       tags: true
+jobs:
+  include:
+    - script: npm run test-client-adapters
+    - script: npm run test-inflection-tables
+    - script: npm run test-components
+    - script: npm run test-data-models
+    # - script: npm run test-res-client
+    - script: npm run test-wordlist
```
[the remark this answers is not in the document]
irina060981: >I was to switch off res-client tests - because for some resons jest failed down after all 4 tests were completely done. I believe that this repo would be anyway refactored - so I switched them off for now.

While I was putting screenshot here - I found a misspell inside in res-client package.json
now it is tested well.
But there are two new problems, I believe from treebank changes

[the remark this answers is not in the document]
kirlat: Oh, I see, that's because the tests work failing. Sorry for missing that comment.

package.json
irina060981: It was defined this way before
you could see that in `npm run test` has all no-coverage format
It was not my change, I believe that you, @kirlat or @balmas, defined to do them without coverage - I don't know why.

May be there are some reasons for that?

packages/components/tests/inflections/inflections-browser-1.test.js
irina060981: will remove - thanks for catch :)

irina060981: Fixed

package.json
kirlat: I've run a history check and it shows that it was added in  109689fe0f0bd8ba8e19cb534488b784a1de6453 as `"test-data-models": "npm --prefix packages/data-models run test"` (if to take data models as an example) and then it was changed in your commit 9c75783295fd8788a82e798af15760f788e04f5c to `"test-data-models": "npm --prefix packages/data-models run test-no-coverage"`. That's if it tells the truth.

.travis.yml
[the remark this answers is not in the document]
kirlat: > But there are two new problems, I believe from treebank changes

What are those tests that are failing? Tests in `res-client` works fine for me. If something fails, we can probably fix it in a separate PR.

package.json
irina060981: If what is me - than it was not by some reason :)
may be I copied from some place and didn't pay attention to it
and then it was reproduced for everyone.

Anyway - if we want to see coveralls - I could easily change it, of course :)

kirlat:
One more question: I've noticed that you're using `it` instead of `test` in Jest files. I'm not familiar with it and I did some search but did not find any significant difference between the two. Wondering which one shall we use in the future. I've read that some developers use `test` for unit testing and `it` for integration testing but I don't know if it's the case here and what should be our best practices regarding that. What do you think?

package.json
balmas: I guess we might as well include the coverage

packages/components/package.json
```diff
@@ -5,8 +5,8 @@
   "main": "dist/alpheios-components.min.js",
   "scripts": {
     "test": "jest --coverage  && shx cat ./coverage/lcov.info | coveralls",
-    "test-no-coverage": "jest tests/lib && jest tests/vue",
-    "test-ci": "npm run test-lib && npm run test-components",
+    "test-no-coverage": "jest tests/lib && jest tests/vue  && jest tests/inflections",
+    "test-ci": "npm run test-lib && npm run test-components ",
```
balmas: so we're testing only vue components when travis runs? Is this what we want?

packages/components/report.20200330.205136.6288.0.001.json
```diff
@@ -0,0 +1,553 @@
+
```
balmas: I think we should exclude the reports from git, unless there is a reason to keep them?

balmas:
> One more question: I've noticed that you're using `it` instead of `test` in Jest files. I'm not familiar with it and I did some search but did not find any significant difference between the two. Wondering which one shall we use in the future. I've read that some developers use `test` for unit testing and `it` for integration testing but I don't know if it's the case here and what should be our best practices regarding that. What do you think?

I think they do exactly the same thing, and it's just a style thing.  I agree standardizing one or the other would be a good idea, don't care at all which.

packages/components/package.json
irina060981: No - this way we test all tests
tests/lib - not vue components
tests/vue - vue components
tests/inflections - full inflection browser tests

packages/components/report.20200330.205136.6288.0.001.json
irina060981: will remove

irina060981:
I mostly use `it` everywhere, from my point view they are aliases 
https://jestjs.io/docs/en/api#testname-fn-timeout

I don't think that trying to divide where it is better to use `it` and where `test` is redundantly.
We won't have any practical benefits but would to spend additional time for deciding and for arguing with each other about it.

Of course it is only my opinion :)

kirlat:
> 
> 
> I will merge it with one failed test
> FAIL tests/lib/selection/media/html-selector.test.js
> 
> ```
>   ● html-selector.test.js › 5 HTMLSelector - setDataAttributes method adds treebank properties to the selection if an element has data-alpheios_tb_src properties and data-alpheios_tb_ref
>     expect(received).toBeDefined()
>     Received: undefined
>       142 |     htmlSel.setDataAttributes()
>       143 | 
>     > 144 |     expect(htmlSel.data.treebank).toBeDefined()
>           |                                   ^
>       145 |     expect(htmlSel.data.treebank.word.sourceUrl).toEqual(alpheios_tb_src)
>       146 |     expect(htmlSel.data.treebank.word.reference).toEqual(alpheios_tb_ref)
>       147 | 
>       at Object.<anonymous> (tests/lib/selection/media/html-selector.test.js:144:35)
> ```
> 
> other review comments are solved

I've fixed that one.

PR #219, opened 2020-03-30
Arethuse treebank update

kirlat:
On top of other changes discussed before this builds adds:
* A reset URL method of updating a treebank iFrame content
* An `experimentalResetTreebankURL` flag to UI Controller's options that triggers the use of the aforementioned method in `treebank.vue`
* Added a `refreshUntilLoaded()` method that will send a `refreshView` until a non-error response is received or a predefined number of attempts is exceeded.

I've fixed all major tests that were broken, but skipped a couple that were very time consuming (related mostly to the LexicalLookupQuery that was removed). Will fix them tomorrow. 

Some tests of inflection tables may still fail. I think this can be due to network issues described by Irina. I also hope on Irina's fixes of inflection tables related tests.


kirlat:
Not sure what to do with those LICENSE.txt files. Should they be ever committed?

packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -284,7 +274,13 @@ if you want to create a different configuration of a UI controller.
...
       // if true, the help button on the toolbar can be controlled by the client, no click handler will
       // be added by the components library
-      overrideHelp: false
+      overrideHelp: false,
+      /*
+      This is an experimental flag that will force treebank iframe URL to be reset every time a treebank tab is
+      opened. It will force a content of an iframe to be reloaded and thus will solve display issues
+      with a treebank view.
+       */
+      experimentalResetTreebankURL: false
```
balmas: I have made the default false because it is working well with the Arethusa fix in place.

packages/client-adapters/src/transformers/alpheios-lexicon-transformer.js
```diff
@@ -188,35 +224,37 @@ class TransformAdapter {
...
-          inflection.addFeature(new Feature(Feature.types.fullForm, targetWord, mappingData.model.languageID))
+          inflection.addFeature(new Feature(Feature.types.fullForm, targetWord, this.mappingData.model.languageID))
         }
         // Parse whatever grammatical features we're interested in and are provided
-        for (const f of this.config.featuresArrayAll) {
+        for (const f of featuresArrayAll) {
           try {
-            mappingData.mapFeature(inflection, inflectionJSON, ...f, this.config.allowUnknownValues)
-            mappingData.overrideInflectionFeatureIfRequired(Feature.types[f[1]], inflection, lemmas)
+            this.mappingData.mapFeature(inflection, inflectionJSON, ...f, this.allowUnknownValues)
```
balmas: fixed a merge error here - it had this.config.allowUnknownValues which which was preventing the inflection data from coming through

packages/data-models/src/treebank_data_item.js
```diff
@@ -0,0 +1,69 @@
...
+        throw new Error('Unsupported treebank application. This version of Alpheios only supports the perseids-treebank-template viewer app.')
+      }
+      if (tbSrcElem.dataset.alpheios_tb_app_version) {
+        this.version = Number.parseInt(tbSrcElem.dataset.alpheios_tb_app_version, 10)
+        if (Number.isNaN(this.version)) { throw new Error(`Treebank version is incorrect in: ${tbSrcElem.outerHTML}`) }
+      }
+
+      if (!tbSrcElem.dataset.alpheios_tb_app_url) { throw new Error(`Missing treebank source URL in: ${tbSrcElem.outerHTML}`) }
+      this.sourceUrl = tbSrcElem.dataset.alpheios_tb_app_url
+
```
balmas: as we discussed, when we eliminated support for the old treebank code, I wanted to use new attributes so that a page that was setup to work with the prior Alpheios release could be updated ahead of the Alpheios release, so that it could work with both the prior release of Alpheios and this one.

So for now I have kept the data-alpheios_tb_ref attribute, whose syntax has not changed and created these new attributes only understood by this version of the code

data-alpheios_tb_app
data-alpheios_tb_app_version
data-alpheios_tb_app_url

We can tweak these later if we need to.

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -77,6 +150,26 @@ Lexis.store = (moduleInstance) => {
...
+       * @param {TreebankDataItem} treebankDataItem - A treebank data item element
+       */
+      setTreebankInfo (state, treebankDataItem) {
+        state.hasTreebankData = treebankDataItem.hasTreebankData
+        state.treebankSrc = treebankDataItem.fullUrl
+      },
+
+      resetTreebankInfo (state) {
+        state.hasTreebankData = false
+        state.treebankSrc = null
```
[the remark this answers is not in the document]
kirlat: I think the next lookup will re-parse the page during creation of a `TreebankDataItem` object. If successful, it will cal `setTreebankInfo` that will set the `treebankSrc` URL. But that might not be fast enough to run queries against it (but we could add a `refreshUntilLoaded()` call before a query to accommodate for that). 

It might be more efficient not to reset a URL every time a lookup is started but rather use a separate variable that will disable treebank queries if a lookup without a context is used, and change a URL only if source changes. Probalby a lot depends on how the treebank app works internally.

packages/client-adapters/src/transformers/alpheios-lexicon-transformer.js
kirlat: Thank you so much for fixing this! I was scratching my head thinking what it might be...

packages/data-models/src/treebank_data_item.js
kirlat: Great, thanks for letting me know!

packages/components/src/vue/vuex-modules/data/lexis.js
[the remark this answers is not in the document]
balmas: >It might be more efficient not to reset a URL every time a lookup is started but rather use a separate variable that will disable treebank queries if a lookup without a context is used, and change a URL only if source changes. Probalby a lot depends on how the treebank app works internally.

Yes, that's what I'm thinking. We need to do some more work here to track the state of the lookups vs the state of the document loaded in Arethusa.  I will try to describe this more clearly in a separate issue.

Issue #220, opened 2020-03-31
endings highlighted in the browser


monzug:
while testing the Greek pronouns, I realized that the endings are still highlighted after few lookups. 
is this an acceptable behavior? 
see screenshot
![pronoun-highlighted](https://user-images.githubusercontent.com/41396793/78027573-993e3400-732b-11ea-8181-cba832bfe949.JPG)

balmas:
no... the state needs to e cleared between lookups.  this is a bug. Thank you!

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200410462

monzug:
the endings are not longer highlighted in the inflection tables browser

PR #223, opened 2020-04-01
Fix classes for cells in inflection browser tables

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/220


irina060981:
there are some problems with tests from parts I didn't touch, may be it is again because of delays in network.
If it is useful, I could spend some time to investigate tomorrow.

packages/components/tests/lib/selection/media/html-selector.test.js
```diff
@@ -131,7 +131,7 @@ describe('html-selector.test.js', () => {
     expect(range).toBeDefined()
   })
 
-  it('5 HTMLSelector - setDataAttributes method adds treebank properties to the selection if an element has data-alpheios_tb_src properties and data-alpheios_tb_ref', () => {
+  it.skip('5 HTMLSelector - setDataAttributes method adds treebank properties to the selection if an element has data-alpheios_tb_src properties and data-alpheios_tb_ref', () => {
```
balmas, line 134: this test is probably broken by changes in #219 . Will enter an issue to fix.

Issue #169, opened 2020-03-02
double-click handler not working on google docs anymore


balmas:
It looks like Google Docs has changed its UI and now our Alpheios double-click handler isn't functioning anymore at all.

the mouseover handler *does* work however, so maybe we could detect if the user is on a google docs site and if so switch automatically to mouseover?

balmas:
dependent upon #183  so that we can notify the users appropriately of the change in behavior

irina060981:
Merged and closed

balmas:
in the QA build with this change (qa.20200320360) the google docs test always returns positive no matter what page you're on.

balmas:
Alpheios Components 3.3.0-qa.20200320403

monzug:
I do get the message that mouse move event is enabled. however the lookup works funny. I did enter a greek word e.g. βουλεύω and initially it did generate the morphology pop-up but then it went back to the lexical query produced no results. have you seen this?

![Screen Shot 2020-03-23 at 5 47 27 PM](https://user-images.githubusercontent.com/41396793/77341621-45ca5580-6d2f-11ea-9ba9-6f5cf9031415.png)

balmas:
hmm. no, I can't reproduce that. @irina060981 can you?

I know we do have more work to do to make the mouseover handler work smoothly, particularly as pertains to the other alpheios ui elements (e.g. mouseover on the panel or popup itself shouldn't trigger a lookup).  See also #144

irina060981:
Yes, I could reproduce it - it fires from button, button label Lookup for me.
After I spent some time for testing, I could see that data-ignore rule doesn't work any more.
I tried to add it directly to label, button - and it gets exclude flag , but lookup still executes.

irina060981:
So it needs more detailed and deep investigation, I could do this, or could first work with Greek input.
@balmas , what do you think?

balmas:
Hi @irina060981 thanks. I think this one is higher priority than the greek input.

irina060981:
ok - will work on it tommorrow :)

irina060981:
By the way it also works on words from all google menus.
And we are not able to add here `alpheios-enabled`.
May be we should limit to google docs id - `docs-editor-container`, where we have our text?

irina060981:
Also this problem is also reproducable on any site if we would turn on mouseover.

balmas:
> By the way it also works on words from all google menus.
> And we are not able to add here `alpheios-enabled`.
[...]

Issue #217, opened 2020-03-27
use fixture for lemma translations test


balmas:
it's failing on travis:

```
 ● lexicons/adapter.test.js › 4 AlpheiosLemmaTranslationsAdapter - getAvailableResLang check availablity of translations
    TypeError: Cannot read property 'forEach' of undefined
      92 | 
      93 |       let mapLangUri = {} // eslint-disable-line prefer-const
    > 94 |       unparsed.forEach(function (langItem) {
         |                ^
      95 |         mapLangUri[langItem.lang] = langItem.uri
      96 |       })
      97 | 
      at AlpheiosLemmaTranslationsAdapter._callee2$ (src/adapters/translations/adapter.js:94:16)
      at tryCatch (../../node_modules/regenerator-runtime/runtime.js:45:40)
      at Generator.invoke [as _invoke] (../../node_modules/regenerator-runtime/runtime.js:274:22)
      at Generator.prototype.<computed> [as next] (../../node_modules/regenerator-runtime/runtime.js:97:21)
      at asyncGeneratorStep (../../node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (../../node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)

```

irina060981:
Merged and closed

PR #229, opened 2020-04-04
Google docs mouse move bugs

irina060981:
For the issue #169 

1. Fixed bug with ignoring `data-alpeios-ignore = all`

2. Fix bug with selecting text inside `alpheios-toolbar`

3. Added options for limited by Id

image


irina060981:
@balmas , I have added here small change for lemma translations
It is in a separate package - so it would be easily to review
And it is a simple addon to the code 

For the issue https://github.com/alpheios-project/alpheios-core/issues/217

balmas:
> Irina I think this is a good start. In reviewing it I realize that I probably need to expand upon the requirements:
> 
> 1. I think maybe we need to go a little further with the configurability of this.  We have to tie to the id to the site
> 2. while we're at it, we should probably have a switch which says whether or not to automatically switch to mouse move for the site.
> 3. we should probably allow for a css selector string to serve as the limit, and not require it to be an element id
> 
> e.g would then have a list of setting pairs;
> 
> Mousemove On:
> 
> site: [ url ]
> limit to: [ css selector ]
> 
> What do you think?

Per discussion, I will turn this request into a new issue :-)

packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -1920,7 +1920,9 @@ NB this is Prototype functionality
     if (this.enableMouseMoveEvent()) {
       const eventParams = {
         mouseMoveDelay: this.featureOptions.items.mouseMoveDelay.currentValue,
-        mouseMoveAccuracy: this.featureOptions.items.mouseMoveAccuracy.currentValue
+        mouseMoveAccuracy: this.featureOptions.items.mouseMoveAccuracy.currentValue,
+        enableMouseMoveLimitedByIdCheck: this.featureOptions.items.enableMouseMoveLimitedByIdCheck.currentValue,
```
balmas, line 1924: @irina060981 I am still thinking that this would apply to ANY mousemove event, not just one on Google docs. Are you sure that it will only be limited by id for the google docs page? enableMouseMoveEvent will return true if mouse move is enabled and the platform is not google docs...

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -197,6 +197,26 @@
       "minValue": 1,
       "maxValue": 5000,
       "values": []
+    },
+    "enableMouseMoveLimitedByIdCheck": {
+      "defaultValue": false,
+      "labelText": "Limit by Id (recommended for Google Docs)",
```
balmas, line 203: can we make this simpler given that it should be only applied to google docs for now?

E.g. maybe a single setting: 

enableMouseMoveOnGoogleDocsId: docs-editor-container

And then we can use this setting both to determine whether or not to auto-enable mouse-move on google docs, and if so, on which element?

balmas: And then the text displayed to the user for the setting could be:

Enable Mouse Move on Google Docs Element Id:

packages/components/src/lib/controllers/ui-controller.js
irina060981: yes, you are right! I am sorry, I thoght how to turn it to more wide variants in future and left it this way.
I will limit to google docs

packages/components/src/settings/feature-options-defaults.json
irina060981: ok, I would change it - but this way id name won't be confugarable (per requirements)

irina060981: @balmas, I believe that our requirements became circual :)
Let's check final requirements

1. We should enable mouseMove on google docs by default
2. We should have ability to limit by id (for now it is fixied to `docs-editor-container`) - via checkbox

irina060981: image

packages/components/src/lib/custom-pointer-events/mouse-move.js
```diff
@@ -11,6 +12,8 @@ export default class MouseMove extends PointerEvt {
     this.boundListener = this.eventListener.bind(this) // A bound event listener of MouseDblClick
     this.mouseMoveDelay = parseInt(eventParams.mouseMoveDelay)
     this.mouseMoveAccuracy = parseInt(eventParams.mouseMoveAccuracy)
+
+    this.limitedById = Platform.getIsGoogleDocs() && eventParams.enableMouseMoveLimitedByIdCheck ? eventParams.mouseMoveLimitedById : null
```
irina060981, line 16: Here it is limited by googleDocsOnly

packages/components/src/vue/components/feature-settings.vue
```diff
@@ -72,6 +72,12 @@
           @change="featureOptionChanged"
       >
       </setting>
+      <setting
+          class="alpheios-feature-options__item"
+          :data="featureOptions.items.enableMouseMoveLimitedByIdCheck"
+          @change="featureOptionChanged"
+      >
+      </setting>
```
irina060981, line 80: I removed the second parameter with id selector only from publishing here, but left it with default value in "private" mode.

packages/components/src/settings/feature-options-defaults.json
balmas: > @balmas, I believe that our requirements became circual :)
> Let's check final requirements
> 
> 1. We should enable mouseMove on google docs by default
> 2. We should have ability to limit by id (for now it is fixied to `docs-editor-container`) - via checkbox
> 3. We could enable mouseMove on any site (it was done before)

oh dear, I think my language was confusing here.

I meant for the setting value to be the id of the element it should be limited to

i.e. default setting value would be "docs-editor-container"

And it would be displayed as such, rather than as a checkbox toggle.

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -197,6 +197,26 @@
...
+        {
+          "value": false,
+          "text": "No"
+        }
+      ]
+    },
+    "mouseMoveLimitedById": {
+      "defaultValue": "docs-editor-container",
+      "labelText": "Limited by the id",
+      "text": true
```
balmas, line 219: so, per my comment in the conversation, I was thinking we could have just one setting, e.g.

```
  "googleDocsMouseMoveLimitedById": {
      "defaultValue": "docs-editor-container",
      "labelText": "Enable Mouse Move on Google Docs Element Id",
      "text": true
}
```

And to fine tune it we could use logic like:

If this wasn't set, then the auto-enable on google docs could be off, 
if it is set to "*" it would mean enable on all elements on google docs
If it is set to any other string it would mean enable on elements with that id on google docs

What do you think?

packages/components/src/vue/components/feature-settings.vue
balmas: see above

packages/components/src/settings/feature-options-defaults.json
irina060981: From my point of view - it is redundantly, because:

1. lokup click event doesn't work properly on google docs, so if we won't enable it by default (as suggested in the issue) - it could confuse users. 
2. it is not clear for most person to use * or `docs-editor-container` to get it work.

I believe that we think about different types of users :)
I thought that we want to help those persons who :
   - won't guess to enable mouseMove manually in settings
   - would confuse when google docs menu would close popup with ready words and would execute lookup on menu titles

I suggest even turn on `enableMouseMoveLimitedByIdCheck` by default.
This way we would protect less It-known persons and give other the ability to tune it (that's why I suggested before to print setting for id name). 

Of course if you think that your suggestion is better - I would implement it :)

balmas: ok, we can keep the two settings if you think it's important. I think we should make these further changes then:

1) mouseMoveLimitedById unprivate - i.e. so that users can change it if they need to (e.g. if this changes on google docs before we can get a new release with a new default value out)

2) change the labelText for enableMouseMoveLimitedByIdCheck to "Auto-enable Mouse Move on Google Docs"

3) change the labelText for mouseMoveLimitedById to "Google Docs Container Id"

I also still think that it would be better if both of these setting names explicitly referred to google docs. e.g. I would use the following names: enableMouseMoveOnGoogleDocs and googleDocsContainerId

irina060981: >change the labelText for enableMouseMoveLimitedByIdCheck to "Auto-enable Mouse Move on Google Docs"

Bridget, I didn't finally understand - does it mean that mouseMove shouldn't be enabled without this check?
and there is no check that defines that we limit by any selector? we should limit, if the mouseMoveLimitedById has value?

balmas: >Bridget, I didn't finally understand - does it mean that mouseMove shouldn't be enabled without this check?

Ok, sorry, I see that I am confusing things. The code currently ALWAYS auto-enables mousemove on google docs and this setting currently controls only if the mousemove will be limited to id, right?

If so, then let's keep that as it is, and change the labelText to 

"Limit Mouse Move on Google Docs to top-level container"

I think we need some usability testing on this generally, so I will stop sending us in circles until we have that. Sorry!

irina060981: > The code currently ALWAYS auto-enables mousemove on google docs and this setting currently controls only if the mousemove will be limited to id, right?

Yes, it works this way:

```
enableMouseMoveEvent () {
    return this.platform.isDesktop && 
              (this.featureOptions.items.enableMouseMove.currentValue || 
               this.options.enableMouseMoveOverride || this.platform.isGoogleDocs)
  }
```

We enable mouseMove in the following case (only for Desktop):
   - enableMouseMove is checked in options (was before)
   - we are on googleDocs

when MouseMove is registered for the listener, it defines limitedBy this way:
```
this.limitedById = Platform.getIsGoogleDocs() && 
                             eventParams.enableMouseMoveLimitedByIdCheck ? 
                             eventParams.mouseMoveLimitedById : null
```

so limited by Id check would work only when it is googleDocs and limitedByIdCheck 

So that's why - @balmas, you understood right!

irina060981:
So finally I am happy to merge it :)

Issue #140, opened 2020-01-15
Application wide configuration


kirlat:
There is a need to have a configuration that will be read by all components of the Alpheios application. One of the examples of such config data can be a URL of a Lexis CS service. This URL is used currently by client adapters and components, and we might need to specify different URLs for dev, QA, and prod environments.

It would be also important to be able to change (reload) configuration without rebuilding an Alpheios application. This might be required, for example, to put a cache and load-balancer in front of the morphology and lexical services, or do other changes within an infrastructure.

Please see an excerpt from a Slack's discussion of the matter:

@balmas: What might be really nice would be to have a url-providing service like a catalog that we could call whenever we need a url for a remote service and we could have all the environments there (dev, qa, prod). I think the more services we add, the more we need this. For example, I'd really like to put a proxy in front of the concordance service (which we don't control), but I can't do that without a release.

@kirlat: The only drawback I see with it is that it can become a single point of failure. But we can probably have a backup configuration within our build or something like that.

@balmas: I think we could address the single point of failure in 2 ways:
(1) having backing defaults in the build config
(2) having redundancy on the service itself
it would be a very simple service to write -- and it's quite possible that there is already something open source that we could just use as-is

@kirlat: We maybe can implement it in the simplest way possible: put a JSON configuration file on a server (as we do with CEDICT data), fetch it and store to IndexedDB or to the local storage. We also could have a matching backup JSON incorporated into our build. If the remote config is not available, we will use a backup JSON from a bundle

@balmas: I like the idea of storing it locally, although we would need to be sure to check for updates

@kirlat: Config file can have a version. Every time Alpheios is loaded it will check if the newer config is available (similar to CEDICT) and will re-download it if necessary.

@balmas: I would also like to leave open the possibility that some urls might be specific to the user or environment. For example, a deployment of the embedded library might want to be able to override certain urls.

@kirlat: So maybe inside this config files we will have different "environments": sets of configuration parameters (dev/QA/production). An Alpheios instance will know which environment it is in and will use the corresponding config data. We can even add a URL params that will specify an environment. That will allow us to switch Alpheios from one mode to the other dynamically. We can also tie other things to the environment. For example, if app is in dev environment, it will print more debug messages and if in production it will not.

[...]

PR #236, opened 2020-04-09
Support for remote configuration

kirlat:
This build adds support for a remote configuration as per #140.


packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -280,7 +282,9 @@ if you want to create a different configuration of a UI controller.
       opened. It will force a content of an iframe to be reloaded and thus will solve display issues
       with a treebank view.
        */
-      experimentalResetTreebankURL: false
+      experimentalResetTreebankURL: false,
+      // A URL of a server that provides an app configuration
+      configServiceUrl: 'https://4wvouc833c.execute-api.us-east-2.amazonaws.com/prod/config'
```
balmas, line 287: this isn't the final url, and probably the url shouldn't be in the code anyway or it should be able to be overridden, maybe via a create argument to the uicontroller. what do you think?

packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -891,6 +905,31 @@ if you want to create a different configuration of a UI controller.
...
+   */
+  async loadAppConfig ({ url, clientId, appName, appVersion, branch, buildNumber } = {}) {
+    if (!url) {
+      throw new Error('An app config server URL is missing')
+    }
+    const configUrl = `${url}?clientId=${encodeURIComponent(clientId)}&appName=${encodeURIComponent(appName)}` +
+      `&appVersion=${encodeURIComponent(appVersion)}&buildBranch=${encodeURIComponent(branch)}` +
+      `&buildNumber=${encodeURIComponent(buildNumber)}`
+    const request = new Request(configUrl)
+    const response = await fetch(request)
```
balmas, line 929: what happens if the request fails?

packages/components/src/lib/controllers/ui-controller.js
balmas: or maybe for now we put it in a default feature options setting?

packages/components/src/lib/controllers/ui-controller.js
balmas: and in the unit tests, will the fetch call be executed or mocked?

packages/components/src/lib/controllers/ui-controller.js
kirlat: > this isn't the final url, and probably the url shouldn't be in the code anyway or it should be able to be overridden, maybe via a create argument to the uicontroller. what do you think?

I've added this as an argument of a UI controller, and a calling app (embed lib or webextension) can override it. Let me submit a PR of a companion `embed-lib` changes that I've used for testing to demonstrate that.

If it's not a final URL then maybe we should not use it as a default value here and use `null` instead of relying on `embed-lib` or `webextension` to set the value? What do you think?

balmas: ah, okay, great. As for default value, let's use https://config.alpheios.net/v1/config 

I will make the necessary terraform config today so that that url will answer.

kirlat: > As for default value, let's use https://config.alpheios.net/v1/config

Great! Updated a default to that value

packages/components/src/lib/controllers/ui-controller.js
kirlat: > what happens if the request fails?

If a request failed, fetch will throw a "Failed to fetch" exception that will fall all the way through. As a result, an embed-lib or webextension will get that exception when they will try to activate a UI controller. I think that's what's desired because thee is no fallback if the remote config server is down. What do you think?

It may only be better to catch a "Failed to fetch" and rethrow as something saying that a remote config cannot be loaded. Would it make it more understandable?

kirlat: > and in the unit tests, will the fetch call be executed or mocked?

Unfortunately, we have a majority of UI controller tests commented out after refactoring. As a result of this, this is not tested currently. I can work on enabling those tests but that will require a serious time and effort.

balmas: ok. I will add an issue for that.

Does the way this is coded prevent that call from being able to be mocked?

balmas: > > what happens if the request fails?
> 
> If a request failed, fetch will throw a "Failed to fetch" exception that will fall all the way through. As a result, an embed-lib or webextension will get that exception when they will try to activate a UI controller. I think that's what's desired because thee is no fallback if the remote config server is down. What do you think?
> 
> It may only be better to catch a "Failed to fetch" and rethrow as something saying that a remote config cannot be loaded. Would it make it more understandable?

I think we should not prevent the UIController from loading if the config fails to load. It should just disable those features that are dependent upon it.

kirlat: > Does the way this is coded prevent that call from being able to be mocked?

Nope. There is a separate function in a UI controller that takes request parameters and returns a config JSON. It could be replaced with a mock easily.

kirlat: > I think we should not prevent the UIController from loading if the config fails to load. It should just disable those features that are dependent upon it.

OK, I will change code to handle that.

balmas: ok thanks!

kirlat:
@balmas, can you please re-review this? I had to make some pretty significant changes to support disabling of CEDICT functionality when LexisCS config is not available. I think it is important not to miss anything there. Thanks!

Issue #84, opened 2020-02-07
restrict the cell matching to the features of the form that produced the table match


monzug:
One of the most common Greek verb is εἶπον
with stemtype aor1 and aor2
Lookup for εἶπον it does highlight too many cells: there should be only the imperative highlighted in the aor1 table and not the indicative 1st person singular and 3rd person plural.
looking at the rules.csv file I did not find anything wrong. am I looking at the right file: inflection-tables/src/paradigm/data/greek/verb/rules.csv ?
tested in Chrome  Alpheios Reading Tools 3.0.6.97
Alpheios Components 1.6.0-dev

monzug:
this is more visible using Irina's tool
see attachments:

1) ἔχω
![Screen Shot 2020-02-12 at 4 59 34 PM](https://user-images.githubusercontent.com/41396793/74352689-27308080-4db9-11ea-972e-4b31c791d4f3.png)

2) εἶπον
![Screen Shot 2020-02-12 at 4 57 49 PM](https://user-images.githubusercontent.com/41396793/74352695-2992da80-4db9-11ea-9d22-86a62e91ac54.png)

monzug:
an other one is ἴω  subj of to go verb.  see screenshot
![Screen Shot 2020-02-13 at 3 31 36 PM](https://user-images.githubusercontent.com/41396793/74445037-05490380-4e76-11ea-96bb-72836c8750b1.png)

monzug:
Note that the release build at texts.alpheios.net, Alpheios Embedded Library 3.1.4.45
Alpheios Components 1.4.7, does not have this problem. it has been introduced recently.
See attachments
![Screen Shot 2020-02-13 at 3 42 10 PM](https://user-images.githubusercontent.com/41396793/74446149-c2882b00-4e77-11ea-8fff-befeeff7e40c.png)
![Screen Shot 2020-02-13 at 3 41 59 PM](https://user-images.githubusercontent.com/41396793/74446155-c451ee80-4e77-11ea-9261-5cf95ebb4f9c.png)

balmas:
ok, this appears to be a bug. We need to restrict the cell matching to the features of the form that produced the table match.

balmas:
@irina060981 I'm not sure how difficult this is to do. Can you assess before undertaking it? Thank you!

irina060981:
It is an easy bug.
There is absent the check if the current inflection matches the current paradigm table (by the rule).
I have fixed - will create a PR.

[...]

PR #244, opened 2020-04-15
Added rule matching check for each inflection before calculate full match

irina060981:
For the issue #84 
image

image


irina060981:
@balmas, there are still some problems with chinese adapter

balmas:
> @balmas, there are still some problems with chinese adapter

ok, I will look at this shortly. thank you!

balmas:
> > @balmas, there are still some problems with chinese adapter
> 
> ok, I will look at this shortly. thank you!

the tests are passing locally. I am not sure what is going wrong with travis but let's see whether it resolves when this PR is merged.

PR #247, opened 2020-04-17
I246

balmas:
adds support for multiple grammars per language, with preferred grammar selected in settings


irina060981:
I have uploaded the branch and test it via embed-lib and there are some inconsistent behaviour here:

If you change grammar in resource tab, the title of grammar will be changes in grammar browser, but grammar content won't not change:

image

and even if I will open/close grammar inside the tab - grammar content won't be updated

irina060981:
Also , I am not sure if it appears with these changes, or may be it was before too.
If you change a page language to not Latin or not Greek language (lookup for not Latin or Greek word) - grammar tab would be completely empty. There won't be any grammar browser.

Looks like this way
image

May be this notice would be helpful

balmas:
Thanks @irina060981 ! I will merge and enter issues for the problems you found. Monica needs to test the index and display of the grammar from within Alpheios so I would like to get a build to her with it included today. Thanks!!

PR #254, opened 2020-04-20
fix voice in greek paradigm 46

### LINKED ISSUE #54: matching problems with verb εἰμί to be:
an other one is related the verb   εἰμί    to be:
1) εἶναι 
2) ἔσεσθαι
the 1) does the match with the εἰμί  inflection table    <stemtype>irreg_mi</stemtype>
2) does the match with   Future System (Active and Middle) with contraction in -έω                                             <stemtype>reg_fut</stemtype> .  (an other example is ἔσεσθον)

this is a situation in which we have the exact verb inflection tables (the verb  εἰμί  table). would it be possible to do the match like we do for the Latin irregular verbs?

_Originally posted by @monzug in https://github.com/alpheios-project/components/issues/846#issuecomment-549869035_

balmas:
fixes #54


balmas:
data fix only, merging without review

PR #255, opened 2020-04-20
fix #83

### LINKED ISSUE #83: extra columns in paradigm tables
1) I did notice that in the Aorist Passive System table we have an extra column not present in Mastronarde tables: 2nd aorist indicative, stem_type aor2_pass
an example of form in the 2nd aorist column (the added column that is not in the paradigm table) is ἐγράφην (lemma γράφω which means scratch, graze)
The other columns matching Mastronarde use lemma βουλεύω which means take counsel, deliberate
shall we just ignore it and delete it from out tables?

2) in the Perfect Active System table we do also have a column for the 2nd perf. ind. such as λέλοιπα (λείπω: leave, quit).
lookup for λέλυκα does highlight also λέλοιπα (1st perfect indicative, deriv_type reg_conj)
<pofs order="1">verb</pofs>
<mood>indicative</mood>
<num>singular</num>
<pers>1st</pers>
<tense>perfect</tense>
<voice>active</voice>
perf_act stemtype

in Smyth, these columns would make the second aorist (active and middle), second perfect and pluperfect active table. see Smyth 368.

I am trying to match Mastronarde tables with Smyth tables but it's not so obvious. will wait for the list of Greek verbs that Vanessa should provide and test the matching.

balmas:
remove stemtype restriction


balmas:
data only fix, merging without review

Issue #200, opened 2020-03-17
add short def to wordlist export


balmas:
(in user target language when we have support for that)

requested by Matteo Romanello

irina060981:
@balmas , would it be enough this way - https://github.com/alpheios-project/alpheios-core/pull/232

```
word	definition
beatum (beatus)	happy, fortunate, bringing happiness; rich, wealthy, copious, sumptuous;
beatum (beatus)	blessed, blissful; "Saint" (in early Church, less formal);
beatum (beo)	bless, make happy, gladden, delight; enrich (with);
beatum (beatus)	happy/fortunate men/persons (pl.); "the_rich"; The Blessed, Saints;
beatum (beatum)	happiness, blessedness; good fortune; good circumstances;
placito (placeo)	please, satisfy, give pleasure to (with dat.);
placito (placitus)	pleasing; agreed upon;
placito (placet)	it is pleasing/satisfying, gives pleasure; is believed/settled/agreed/decided;
placito (placitum)	plea;
placito (placito)	plead;
```

Or should we add all definitions from all lexeme to one line?

balmas:
Ideally, I think we would have each lexeme on its own line, as you have here, and if a lexeme's lemma has principal parts, we should include them, similar to the way we do in the popup. so, in the above example, for beatum, the output would be:

```
`
word	definition
beatum (beatus, beata, beatior, beatissimus)	happy, fortunate, bringing happiness; rich, wealthy, copious, sumptuous;
beatum (beatus, beata, beatum)	blessed, blissful; "Saint" (in early Church, less formal);
beatum (beo, beare, beavi, beatus)	bless, make happy, gladden, delight; enrich (with);
beatum (beatus, beati)	happy/fortunate men/persons (pl.); "the_rich"; The Blessed, Saints;
beatum (beatum, beati)	happiness, blessedness; good fortune; good circumstances;
```
 
This information, as well as the short definition, can be obtained from the full homonym, but we don't store that on the remote server.

Although the user may have a local store of the entire wordlist, they may not, so we have to decide what to do about that. We can either

(1) always store the entire homonym on the remote server 
(2) store only the additional parts we need for quizlet download (i.e. principal parts and short definitions) on the remote server
(3) if a quizlet download has been requested, for any word on the list which doesn't have a full homonym available in the local cache, then we first populate via lookup and then download the list. 
[...]

PR #266, opened 2020-04-24
Wordlist export features

irina060981:
For the issue #200 

1. Added new fields for worditem
    - updatedDT
    - frequency

2. Added additional filter
    - most recent (by updatedDT)
    - most often (by frequency)

    
image

3. each additional filter has filterAmount field, it defines amount of words in filter

image

    it gets default data from `featureOptioins.wordlistFilterAmountDefault`

4. added another `featureOptions.wordlistMaxFlashcardExport` option
    used inside download confirmation block
    is saved to user profile settings if changed

    
image

5. On flashcard export if worditem doesn't have shortDefs - it inits additional lexicalQuery
    and download the wordlist only after all lexical requests are complete
    while lexical data is loading - progressbar is shown
    
image

6. also added `frequency` and `updatedDT` to wordlist ui and added sorting for them


irina060981:
will update tests later, forgot about them
anyway - @balmas , would be glad to get some feedback!

balmas:
@irina060981 I think this is a great start ! Thanks for trying to make sense of the requirements :-)

a few thoughts on the UI:

(1) I don't think I want it to be so easy for the user to change the max items for download, since this something we are restricting to prevent performance issues.  So rather than offering them the option to change the max on the download dialog, I'd like to just show them a notice, e.g. something like "Note: the Flashcard export is limited to a maximum of XX words." where XX is taken from the setting.  And then we can put the option to change the setting on the Features option tab.

(2) I'm thinking about these 2 levels of limits and I think they are a little confusing and I'm not sure it would make sense for them to be used in conjunction.  Can we just put the new limits in the existing box? And then precede the input box for the `wordlistFilterAmountDefault` value with the word "Top" ? (I expect we will have to have a few iterations on the UI but I think I'd like to start with that).

(3) the a/z sort icons need to be different for the new fields, because the alphas don't make sense there. Can we just have up/down arrows?

(4) with the additional fields, we probably need to add column headers to the display. I don't think it's obvious what the number and the date mean.

Can you make those changes and then I'll take a closer look at the code?

Thanks!

irina060981:
Updated ui
image

image

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -200,7 +200,10 @@ export default class Lexis extends Module {
     }
     if (!wordUsageExamples) { wordUsageExamples = this._appApi.getWordUsageExamplesQueryParams(textSelector) }
 
-    this._appApi.newLexicalRequest(textSelector.normalizedText, textSelector.languageID, textSelector.data, source)
+    if (!forWordlist) {
+      this._appApi.newLexicalRequest(textSelector.normalizedText, textSelector.languageID, textSelector.data, source)
+    }
```
balmas, line 205: I'm wondering if there are going to be any unexpected side effects of issuing the lexicalQuery.getData request without going through the regular process of initializing a new lexical request. getData issues a number of events that affect the interface.  It might be that we need to do more here, and suppress events during the lexical query getData when we're doing this sort of background request. I'd like to get some thoughts from @kirlat on this.

packages/components/tests/vue/components/word-list/word-language-panel.test.js
```diff
@@ -594,7 +608,7 @@ describe('word-language-panel.test.js', () => {
     expect(result[2].targetWord).toEqual('cupidinibus')
   })
 
-  it('26 WordLanguagePanel - method downloadList executes Download.collectionToCSV, Download.downloadBlob and hides downloadAll confirmation box', () => {
+  it.skip('26 WordLanguagePanel - method downloadList executes Download.collectionToCSV, Download.downloadBlob and hides downloadAll confirmation box', () => {
```
balmas, line 611: why are we skipping this test now?

packages/wordlist/src/controllers/user-data-manager.js
```diff
@@ -110,6 +110,7 @@ export default class UserDataManager {
     }
     try {
       params.source = params.source||'both'
+      // params.source = 'local'
```
balmas, line 113: remove this and below?

packages/wordlist/src/storage/remote-db-adapter.js
```diff
@@ -172,6 +171,9 @@ export default class RemoteDBAdapter {
           this.errors.push(error)
         }
       }
+      if (error.message === 'Request failed with status code 401') {
+        return []
+      }
```
balmas, line 176: why the change in behavior here?

packages/wordlist/src/storage/worditem-remotedb-driver.js
```diff
@@ -183,6 +186,8 @@ export default class WordItemRemoteDbDriver {
     } else {
       result.context = []
     }
+    // console.info('remote _serialize wordItem', wordItem.frequency, wordItem)
```
balmas, line 189: remove debugging

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -233,6 +233,22 @@
         }
       ],
       "limitByLangs": ["lat", "grc"]
+    },
+    "wordlistMaxFlashcardExport": {
+      "defaultValue": 25,
+      "labelText": "Wordlist Flashcard Max export",
+      "number": true,
+      "minValue": 1,
+      "maxValue": 5000,
```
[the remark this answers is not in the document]
irina060981: fixed

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -233,6 +233,22 @@
...
+      "values": []
+    },
+    "wordlistFilterAmountDefault": {
+      "defaultValue": 10,
+      "labelText": "Wordlist Filter Amount default",
```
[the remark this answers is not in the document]
irina060981: done

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -233,6 +233,22 @@
...
+      "number": true,
+      "minValue": 1,
+      "maxValue": 25,
```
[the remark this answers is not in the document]
irina060981: done

packages/components/src/locales/en-us/messages-word-list.json
```diff
@@ -144,5 +144,26 @@
...
         "description": "Download list for flashcards",
         "component": "WordLanguagePanel"
+    },
+    "WORDLIST_FILTER_BYMOSTRECENT": {
+        "message": "by most recent",
+        "description": "Filter by max amount with sorting by updatedDT desc",
+        "component": "WordFilterPanel"
+    },
+    "WORDLIST_FILTER_AMOUNT_PLACEHOLDER": {
+        "message": "top amount",
```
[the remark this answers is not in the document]
irina060981: done

packages/components/src/vue/components/word-list/download-confirmation.vue
```diff
@@ -0,0 +1,229 @@
...
+        Download.downloadBlob(result, `wordlist-${this.languageCode}.csv`)
+        this.$emit('changeShowDownloadBox', false)
+      },
+      cancelDownloadList () {
+        this.$emit('changeShowDownloadBox', false)
+      },
+      featureOptionChanged (name, value) {
+        let keyinfo = Options.parseKey(name)
+
+        console.info('featureOptionChanged - ', keyinfo.name, value)
```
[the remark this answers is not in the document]
irina060981: done

packages/components/src/settings/feature-options-defaults.json
```diff
@@ -233,6 +233,22 @@
...
+    "wordlistMaxFlashcardExport": {
+      "defaultValue": 25,
+      "labelText": "Wordlist Flashcard Max export",
+      "number": true,
+      "minValue": 1,
+      "maxValue": 5000,
+      "values": []
+    },
+    "wordlistFilterAmountDefault": {
+      "defaultValue": 10,
```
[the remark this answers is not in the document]
irina060981: done

packages/components/src/vue/components/word-list/word-filter-panel.vue
```diff
@@ -2,50 +2,64 @@
...
-            <input v-model="textInput"  class="alpheios-input alpheios-wordlist-header-input-filterBy"
-                  :placeholder="currentTypeFilter.textInputPlaceholder"
-                  v-on:keyup.enter = "clickFilterBy"
-                  v-on:input = "filterVariants"
-                  v-on:focus = "filterVariants"
-                  v-on:blur = "hideAutocomplete"
-                  autocapitalize = "off"
-                  autocorrect = "off"
-                  >
+          <div class="alpheios-wordlist-header-select-filterBy-first">
```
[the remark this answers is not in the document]
irina060981: no, it is a wrapper around the number filed - I have changed the class name to reduce misunderstanding

packages/components/src/vue/vuex-modules/data/lexis.js
irina060981: From my point of view `newLexicalRequest` method change data in ui, app, storage modules.
It is not needed for background process of updating homonym.
I belive that we need only to update homonym object here - that's why I think it is ok.
Would be glad to @kirlat 's opinion.

packages/components/tests/vue/components/word-list/word-language-panel.test.js
irina060981: I need to update it because this features was moved to a separate component - fixed

packages/wordlist/src/controllers/user-data-manager.js
irina060981: removed

packages/wordlist/src/storage/remote-db-adapter.js
irina060981: oh, I added it after problems with testUser - without this it was failed with unproper result
I belive that it is better to simply get empty array and of course show all errors as before including this 401 error

packages/wordlist/src/storage/worditem-remotedb-driver.js
irina060981: done

irina060981:
@balmas , if I understood you right - I could merge it after above small fixes.
will do that :)

packages/components/src/vue/vuex-modules/data/lexis.js
kirlat: @irina060981 I've tried to simplify a number of options of lexical query methods and I think it is important in order to keep it maintainable. Right now we have two types of requests identified by a `source` prop: a get selected text request from a page (identified by a `page` source) and a one initiated by the lookup component (source is `lookup`). It looks to me (please correct me if I'm wrong) that this is a third type of request. So maybe we can use a `source` filed to identify it (i.e. having it set to something like `wordList`)? We can modify a `lookupText()` API method so that it will take `source` as an additional parameter and will pass it through to `lexicalQuery`. Or we can create another special method like `wordListLookup()` that will be similar to `lookupText()` but will use a source of `wordList`. What do you think about that?

packages/wordlist/src/storage/remote-db-adapter.js
balmas: ok, thanks!

packages/components/src/vue/components/word-list/word-filter-panel.vue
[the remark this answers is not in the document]
balmas: ok thanks.

Issue #96, opened 2018-07-23
selective enabling of Alpheios on components


balmas:
for alpheios-project/components#129 we disabled alpheios on the panel and popup . sometimes we want to be able to enable it selectively. This requires some thought about the best way to do it but, for example, the ge'ez parser provides short definitions for its words in Latin. We should able to enable Alpheios on those definitions in the popup.

One way to do this might be to look for the language code on the text that is displayed in a component and compare that to available languages to determine if Alpheios can be activated. But we might need even finer grained control of that.

Probably an issue for after refactoring of component state and data.

balmas:
I need to write clear requirements for this before we can work on it.

balmas:
I have done the following:

(1) added support for a new data attribute `data-alpheios-enabled="all"` . If this attribute is set in a child of an element which has the `data-alpheiosignore="all"` attribute set, then it will override the ignore instruction and enable alpheios for the element and its children. NB: this should be documented in the FAQ and embed-lib documentation. I added a documentation label to issue to remind us to do that...

(2) I have made a chance to the short definition component on the popup. If the definition is in a language that is supported by Alpheios (i.e one of the Alpheios target languages) then it will get this data attribute and Alpheios lookups will be enabled on the definition text. 

This can be tested with the Ge'ez word "ሀገርየ"" which has a definition in Latin.  

Additional issues can be entered to cover adding similar functionality on other elements of the Alpheios UI.

``

balmas:
Alpheios Components 3.3.0-qa.20200501441

monzug:
I tested the word ሀገርየ and I was able to lookup terra inhabitata and ager cultus. I have a question, as these lookups are initiated as child of a word in an other language, would it make more sense (or be more practical) to have the initial lookup on the background so when I quick look for the e.g. latin word patria, I can go back to my original word ሀገርየ?

balmas:
> I tested the word ሀገርየ and I was able to lookup terra inhabitata and ager cultus. I have a question, as these lookups are initiated as child of a word in an other language, would it make more sense (or be more practical) to have the initial lookup on the background so when I quick look for the e.g. latin word patria, I can go back to my original word ሀገርየ?

It would but that is a fairly big change. I think we do have already a separate issue for supporting multiple popups #86 . This is a good use case for that issue.

monzug:
an other one: I have tried on ὀδόντων. who has latin words in the short
[...]

PR #270, opened 2020-04-24
wip on #96

balmas:
this PR enables lookups on short definitions if the short definition is in one of the supported Alpheios languages.


balmas:
> Looks good to me! I have some thoughts about the tag format. Initially I was not able to understand what is it doing and got only checking the source code. I think what puzzled me was a double negative in `data-alpheios-ignore="none"`. Maybe it's just me but I'm thinking maybe something positive will be simpler to understand? I.e. something like `data-alpheios-enable="all"`. So with `data-alpheios-ignore` we disable handling of something, and what exactly is disabled is listed in the value field (i.e. "all"). If we'll use `data-alpheios-enable` it would be the exact opposite, and we would specify the same way what is enabled ("all" or some particular actions). What do you think about such approach?
> 
> If we need to override "disable" just for short definitions then we could also do that with Vue click handler, that will be simpler, on my opinion. But if we're going to use it in other places too (and probably we do) then having a data attribute would make sense.
> 
> What do you think?

I agree it's confusing. I chose to reuse data-alpheios-ignore because I was thinking of it as reversing the higher up use.  I don't really know what is easier to understand, but I'm fine with your suggestion so I'll implement that before merging. Thanks!

Issue #221, opened 2020-03-31
manage state between Alpheios and Arethusa


balmas:
This issue follows up on the code review comments at 

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/pull/219_

balmas:
Alpheios requirements:

1. Alpheios must display the tree icon in the toolbar and popup if the current word has treebank data
2. Alpheios must NOT display the tree icon in the toolbar and popup if the current word does not have treebank data
3. Alpheios MAY display the tree icon in the toolbar before the first lookup if the currently loaded page contains treebank references

Arethusa limitations:

1. Arethusa must be fully loaded with a document before it can respond to the getMorph API request
2. Arethusa must have its state set to the current sentence before it can respond to the getMorph api request for a word in that sentence

balmas:
State that needs to be managed:

1. whether the page has any treebank documents
2. the list of treebank documents referenced by the page (currently only 1 supported but we want to support more per #148)
3. the treebank sentence/word reference for the current word
4. the treebank sentence/word reference for the previous word

balmas:
![tree_workflows](https://user-images.githubusercontent.com/644870/78048068-3064b500-7347-11ea-993b-e6d085d1d5ff.png)
![load_arethusa_document](https://user-images.githubusercontent.com/644870/78048075-335fa580-7347-11ea-9c47-407d664063f0.png)

balmas:
@kirlat when you have a minute can you take a look at the above and let me know if it makes sense and if you see any think that raises concerns?

kirlat:
This is a very nice and detailed diagrams, thanks! 

The only question I have is about the use of `show_tree_for_page` variable. In my understanding, the only use for that is to decide whether to show a treebank icon during the initial page load, is that so? Because after the page is loaded and lookups have started it's the last word being looked up determines whether the treebank icon will appear (i.e. whether the word being looked up has any reference or not)? Of course, if the page has no treebank data at all then word lookups do not matter: we will not show the treebank icon no matter what.

Is that correct?

balmas:
[...]

PR #275, opened 2020-04-28
Fixes known issue in #221

kirlat:
Fixes the issue with URL being reset every time a sentence changes. 
Prevent treebank icon from flashing during refresh. 
For #221.

It also fixes some discrepancies in popup content that I've found.


packages/components/src/vue/components/morph-parts/principal-parts.vue
```diff
@@ -40,20 +41,20 @@
             :type="types.pronunciation"
         />
 
-        <span class="feature_extras" v-if="hasExtras" >
+        <div class="feature_extras" v-if="hasExtras" >
```
balmas, line 44: is this changing them from displaying inline to block?

kirlat: Because `span` is an inline element and we cannot use margins for it (which we need to set spacing). We're also using `display: flex` for a parent element, and that works better when its children are block-level elements. I just think `display: flex` is so powerful (can set spacing an alignment of elements in both directions in many different ways) that we should use it whenever possible (Bootstrap recognized that by switching to `dispaly: flex` in its latest version).

PR #277, opened 2020-04-28
fix #16

### LINKED ISSUE #16: numeric greek headword should not result in error
lookup for     χρῆ   - it kept loading and loading and loading

console says: 
Unhandled Promise Rejection: TypeError: e.normalize is not a function. (In 'e.normalize("NFC")', 'e.normalize' is undefined)

balmas:


packages/client-adapters/src/transformers/alpheios-lexicon-transformer.js
```diff
@@ -177,7 +177,9 @@ class AlpheiosLexiconTransformer {
         const index = entry[0]
         const elem = entry[1]
 
-        const lemmaText = elem.hdwd ? elem.hdwd.$ : undefined
+        // if the parser has a number in the hdwd the JSON will have it as a number
+        // and not a String
+        const lemmaText = elem.hdwd ? `${elem.hdwd.$}` : undefined
```
kirlat, line 182: Can we always be sure that the value of `$` is set? Because if `$` is not present in `elem.hdwd` then the value of `${elem.hdwd.$}` will be a string containing "undefined". This is probably not what we want.

So if we're not sure about the presence and value of `$` maybe we can add an extra check? Like `const lemmaText = elem.hdwd && elem.hdwd.$ ? `${elem.hdwd.$}` : undefined`? What do you think?

packages/client-adapters/tests/transformers/alpheios-lexicon-transformer.test.js
```diff
@@ -153,4 +154,22 @@ describe('transform-adapter.test.js', () => {
     expect(homonym.lexemes[0].lemma.word).toEqual('hdwdb')
     expect(homonym.lexemes[0].inflections.length).toEqual(1)
   })
+
+  it('2 TransformAdapter - makes sure Lemmas are strings', async () => {
```
kirlat, line 158: I'm thinking maybe we should also add a test case when `$` is not defined? What do you think?

packages/client-adapters/src/transformers/alpheios-lexicon-transformer.js
balmas: it's a good suggestion. will make this change.

packages/client-adapters/tests/transformers/alpheios-lexicon-transformer.test.js
balmas: yes good idea.

Issue #205, opened 2020-03-23
Moving the pop-up goes partially out of screen and cannot see the top nav


monzug:
I did experience this issue multiple times in the last builds: sometimes (not always) when I move the pop-up on top of the screen, the top nav is not visible and when I move it down the screen and release, it goes again up. pretty annoying. However I do not have a reproducible scenario. Anyone has seen this?

balmas:
yes I think I saw that too. Giving to @irina060981 to debug.

balmas:
reassigned to @kirlat

kirlat:
@balams: Strangely, I cannot reproduce this with the latest version of `alpheios-core` (`i140-app-wide-config`). There was an `interactjs` update there so maybe, hopefully, it fixed that. Or maybe I'm just not following the right steps for it to appear. 

Can you please verify that it still exists for you in this version? Thanks!

balmas:
ok, will ask @monzug to retest with the next build.

balmas:
@monzug can you retest in Alpheios Components 3.3.0-qa.20200410462 ?

monzug:
Just experienced this issue again in Chrome on latest build: Alpheios Reading Tools 3.1.0 build qa.20200420272
Alpheios Components 3.3.0-qa.20200420262
see screenshot
![pop-up-move-up](https://user-images.githubusercontent.com/41396793/79871792-e6557900-83b2-11ea-92c5-9c20e29a9adc.JPG)

to repro: lookup for caro. then go to the bottom of the pop-up and click on Credits. Cannot see the top nav any longer. If I move to the top nav and release, the pop-up will go back to show only the bottom part so I am not able to close the pop-up.

kirlat:
@monzug: thanks for a detailed description! I see the problem now, will fix.

balmas:
Alpheios Components 3.3.0-qa.20200423418

monzug:
tested in Chrome on PC, build Alpheios Reading Tools 3.1.0 build qa.20200423423
Alpheios Components 3.3.0-qa.20200423418
it works well in Loeb classics but not in latin library.
I am still able to repro the above scenario (lookup for caro and expand the credits, then can't reach the top nav)

kirlat:
> it works well in Loeb classics but not in latin library.
> I am still able to repro the above scenario (lookup for caro and expand the credits, then can't reach the top nav)

Hmm, cannot see this in my tests somehow. This is how it looks in my case:
[...]

PR #278, opened 2020-04-28
A fix for a browser viewport height calculation

kirlat:
A fix for a viewport height not being calculated correctly when a browser is in a quirks mode. See #205.


packages/components/src/lib/utility/platform.js
```diff
@@ -32,16 +32,36 @@ export default class Platform {
...
+    const clientHeight = (document.compatMode === 'BackCompat')
+      ? document.body.clientHeight
+      : document.documentElement.clientHeight
+
     this.viewport = {
       width: window.innerWidth && document.documentElement.clientWidth && document.body.clientWidth
         ? Math.min(window.innerWidth, document.documentElement.clientWidth, document.body.clientWidth)
         : document.body.clientWidth || window.innerWidth || document.documentElement.clientWidth,
 
+      // The following should calculate a viewport height correctly in both quirks and non-quirks modes
```
balmas, line 50: but you're still only using documentElement.clientHeight here? why is the clientHeight const only used for the scrollbars?

kirlat: This works, but is not very precise. However, extreme precision was not important for our purpose so I decided not to change it because it is a tested code that is used in several places.

`window.innerHeight` works correctly in both modes. This is a height of a viewport including a scrollbar.

In a "no-quirks" mode `documentElement.clientHeight` is a height of a viewport without a scrollbar. In a "quirks" mode is it a height of a document (a value usually higher than a viewport width). So, depending on the mode and a scrollbar presence `Math.min(window.innerHeight, document.documentElement.clientHeight)` will return a height of a viewport with or without a scrollbar. It is not a problem for the popup height as we have extra margins of 60px and, because scrollbar height is 17px, it will be an extra gap of 17px sometimes, almost not noticeable. It will appear only when a popup is too tall and its height is restricted.

But since I'm touching this probably it will be good to make adjustments here too. Maybe we can have a `viewport`  (or maybe call it outerViewport) to report dimensions including a scrollbar (i.e. innerWidth/innerHeight), then have a scrollbar width props, and then a getters that will return width and height of of a viewport without sidebars (maybe we can call it `innerViewport`? Not sure what's the best name for it)?

What do you think?

balmas: I think that's a good idea. Because I think otherwise we could come back to this at some point and forget why we only applied it in one place. inner and outer viewport sound like ok names to me. thanks.

kirlat: Great, will update it tomorrow.

kirlat:
I've added calculations of viewport dimensions with and without scrollbars. I decided to refer to the "outer" dimension as simply "viewport width/height" because that's consistent with a widely accepted practices.
Dimensions without scrollbars are referred to as "inner width/height". I've added comments explaining meanings of those values.

@balmas: please re-review when you will have time. Thanks!

packages/components/src/lib/utility/platform.js
```diff
@@ -32,14 +32,42 @@ export default class Platform {
...
+        width: window.innerWidth - document.documentElement.clientWidth
+      }
+    }
 
-      height: window.innerHeight && document.documentElement.clientHeight
-        ? Math.min(window.innerHeight, document.documentElement.clientHeight)
-        : window.innerHeight || document.documentElement.clientHeight
+    this.viewport = {
+      // A width of the viewport that includes the width of the scrollbars, if they are shown
+      width: window.innerWidth,
```
balmas, line 56: is there any chance window.innerWidth is not defined? Is it safe to drop the fallbacks? Same question for innerHeight

kirlat: I did not find any information about cases where it might not behave as expected, except for buggy implementations in some very early versions of some browsers that are now not in use now. I also checked it on several sites known to be problematic, and it worked well there.

I will probably merge it now: it will go through a series of tests anyway, and if there will be any issues, we'll notice and will have time to fix them.

balmas: sounds good.

Issue #186, opened 2020-03-09
test greek noun, adjective paradigms


balmas:
these were added via alpheios-project/inflection-tables#289

Can be tested in Alpheios Components 3.3.0-qa.20200309431

monzug:
how do I test this? If I enter a noun in the lookup, I do still get the noun declension or the simplified declension. If I enter a noun in Irina's tool, I do only get Morpheus output. Please advise

balmas:
hmm. Maybe we need to do more work to finish this, I was thinking it was done, but will reopen and investigate. Thanks.

balmas:
@irina060981 can you investigate? Not sure what remains to be done here. Thanks!

irina060981:
My tool could check only verbs and participles - because it extracts only paradigms.
That's why for nouns it shows nothing.
@balmas , @monzug, Do we need to expand the tool for all type of inflection tables?

balmas:
this is not the inflection-input tool, but the actual alpheios application (webextension/embed-lib) that @monzug is testing (I think anyway)

irina060981:
>how do I test this? If I enter a noun in the lookup, I do still get the noun declension or the simplified declension. If I enter a noun in Irina's tool, I do only get Morpheus output. Please advise

@balmas, then I am not sure - what exactly I should investigate here? The described workflow is as it exists now upon the requirements in issues. Or Should I point my attention to anything here?

balmas:
let me take another look. thanks!

balmas:
ok, here's what I think needs to be done still:

1. Add the new noun and adjective paradigms to the inflection browser 
2. Change the logic so that when you look up a noun or adjective that returns a paradigm table, it returns the paradigm table *in addition to* as the regular endings tables (right now it returns only the paradigm table if a paradigm table matched)

Thanks!

irina060981:
You are right, I have forgotten about the fact I added paradigm tables for other part of speech myself :)
Will add them here too :)

irina060981:
Add the new noun and adjective paradigms to the inflection browser
Change the logic so that when you look up a noun or adjective that returns a paradigm table, it returns the paradigm table in addition to as the regular endings tables (right now it returns only the paradigm table if a paradigm table matched)

Both features are released

irina060981:
I believe that I could close it

balmas:
Alpheios Components 3.3.0-qa.20200501441

monzug:
[...]

PR #283, opened 2020-04-29
Add noun and adjective paradigm tables to browser

irina060981:
For the issue #186 

Added noun and adjective paradigms to browser
image

change logic - show paradigm tables for noun and adjective in addition to others
image


packages/inflection-tables/tests/data/greek-adjective-paradigm.test.js
```diff
@@ -31,15 +31,15 @@ describe('greek-adjective-paradigm.test.js', () => {
 
     expect(inflectionsViewSet.hasMatchingViews).toBeTruthy()
 
-    expect(inflectionsViewSet.matchingViews.length).toEqual(2)
+    expect(inflectionsViewSet.matchingViews.length).toBeGreaterThanOrEqual(4)
```
balmas, line 34: why not test for exact equality on number of views?

irina060981: Oh, I only simplified tests for me - but if you think - it is important - I will fix it

irina060981:
@balmas, I have updated all checking inflections tests with strict equality
Now I will merge it

packages/inflection-tables/tests/data/greek-adjective-paradigm.test.js
balmas: thanks! I think it's good to have an exact test because it can help us find when we've done something that changes it.

Issue #251, opened 2020-04-20
in Loeb Classics, the buttons have a black, not white background


monzug:
see screenshot
![missing-greek-verb-inflection](https://user-images.githubusercontent.com/41396793/79762496-da08e780-82f0-11ea-9d77-4a1761ae14b2.JPG)

monzug:
this is in chrome on mac
![Screen Shot 2020-04-22 at 2 13 46 PM](https://user-images.githubusercontent.com/41396793/79980559-94d8e700-84a3-11ea-887f-603eb35a1bf3.png)

monzug:
also, the buttons in the Browse grammar  look off on the right side.  this is in Chrome on pc and mac
see screenshot
![new-latin-grammar](https://user-images.githubusercontent.com/41396793/80099739-f5683280-853c-11ea-88ff-f4749bc8aa4c.JPG)

balmas:
Alpheios Components 3.3.0-qa.20200501441

monzug:
Now we have a white background color but when mouse over Don't ask me again it's still black, see screenshot is #292

kirlat:
> Now we have a white background color but when mouse over Don't ask me again it's still black, see screenshot is #292
> also the two buttons are not aligned as per release

Hope that the latest changes just merged into master would fix this.

balmas:
Alpheios Components 3.3.0-qa.20200513442

monzug:
YES, it looks fine

PR #284, opened 2020-04-29
Style fixes for Loeb Clasics

kirlat:
Fixes an appearance of some buttons on Loeb Classics website, see #251.


kirlat:
> ok if there is no other way besides using !important I'm okay with it

They are using `!important` so we have no choice 🙁

Issue #252, opened 2020-04-20
see declension in the participle table does not work


monzug:
I think we had this issue before. anyhow, in the current build, if I click on see declension it does not do anything. on the contrary in the release, it takes to the related participle table (e.g. βουλεύσομαι)
tested in Chrome in Alpheios Reading Tools 3.1.0 build qa.20200410484
Alpheios Components 3.3.0-qa.20200410462

balmas:
I know we've fixed this before. Maybe related to recent changes for subtables. @irina060981 can you investigate?

irina060981:
I need some additional explanations here -  
I used `βουλεύσομαι` for testing (thank you @monzug for this example)

I have tested several elder branches:
`3.0.3.90-branch-for-release`
`3.0.1.82-branch-for-release`

see declension link doesn't do anything visible

I investigated the current code:
and there was missed `navigate` behaviour.
I added it - and it should scroll to suptable `verbpdgm65` that is right under it -

this way - behaviour is not visible, but I believe that it is right.

@balmas and @monzug , am I right? and in this example there wouldn't be any visible result on clicking `see declension`

irina060981:
You could see changes for `navigate` method in PR

balmas:
>in this example there wouldn't be any visible result on clicking see declension

I'm not sure I understand. The result should be the scrolling to the see table

irina060981:
>I'm not sure I understand. The result should be the scrolling to the see table

as I could see from debug - it searches and scrolls

merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200501441

monzug:
yes, it works in Alpheios Reading Tools 3.1.0 build qa.20200507338
Alpheios Components 3.3.0-qa.20200507327

PR #286, opened 2020-04-30
Navigate behaviour for link in invelections tab

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/252


packages/components/src/vue/components/inflections/inflections-subtables-wide.vue
```diff
@@ -73,4 +73,9 @@ export default {
   .infl-prdgm-tbl__cell--label {
     font-weight: 700;
   }
+
+  .infl-prdgm-tbl a.infl-prdgm-tbl__cell-reflink {
```
balmas, line 77: can we also add these styles to the "Back to Main" link?

irina060981: Fixed styles - also added hover styles for both - sub and supp tables

balmas: thanks!

Issue #271, opened 2020-04-28
Update usage of lexis.lexicalQuery for wordlist upload


irina060981:
From the PR https://github.com/alpheios-project/alpheios-core/pull/266

I'm wondering if there are going to be any unexpected side effects of issuing the lexicalQuery.getData request without going through the regular process of initializing a new lexical request. getData issues a number of events that affect the interface.  It might be that we need to do more here, and suppress events during the lexical query getData when we're doing this sort of background request. I'd like to get some thoughts from @kirlat on this.

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/pull/266_

irina060981:
From my point of view `newLexicalRequest` method change data in ui, app, storage modules.
It is not needed for background process of updating homonym.
I belive that we need only to update homonym object here - that's why I think it is ok.
Would be glad to @kirlat 's opinion.

_Originally posted by @irina060981 in https://github.com/alpheios-project/alpheios-core/pull/266_

irina060981:
@irina060981 I've tried to simplify a number of options of lexical query methods and I think it is important in order to keep it maintainable. Right now we have two types of requests identified by a `source` prop: a get selected text request from a page (identified by a `page` source) and a one initiated by the lookup component (source is `lookup`). It looks to me (please correct me if I'm wrong) that this is a third type of request. So maybe we can use a `source` filed to identify it (i.e. having it set to something like `wordList`)? We can modify a `lookupText()` API method so that it will take `source` as an additional parameter and will pass it through to `lexicalQuery`. Or we can create another special method like `wordListLookup()` that will be similar to `lookupText()` but will use a source of `wordList`. What do you think about that?

_Originally posted by @kirlat in https://github.com/alpheios-project/alpheios-core/pull/266_

irina060981:
as I could see from the code - source is used similiar to my additional variable,
`source` is simply passed to `newLexicalRequest` method, and my variable is simply ignores `newLexicalRequest`

I didn't use source because I was not sure how widely it is used in the code.

I think, that it is useful to create a separate method for page lookup, input lookup and wordlist lookup, because both methods:
  - for input lookup and wordlist lookup doesn't need treebank analysis (because it couldn't have any treebank marks)
  - for wordlist we also don't need any ui events

@kirlat and @balmas , what do you think?

kirlat:
> I think, that it is useful to create a separate method for page lookup, input lookup and wordlist lookup

[...]

PR #287, opened 2020-04-30
Refactor lexis lookup

irina060981:
For the issue #271 

I decided that this refactoring is enough to reduce duplicated code:

- I moved treebank check to a separate metod
- use `source === page` to execute get treebank data 
- use `source  === wordlist` to prevent newLexicalRequest execution
- created simple `lookupForWordlist` similiar to `lookupText`


packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -424,8 +437,19 @@ Lexis.api = (moduleInstance, store) => {
...
      * @param wordUsageExamples
      */
-    lookupText: async (textSelector, forWordlist = false) => {
-      return moduleInstance.lexicalQuery({ store, textSelector, source: 'lookup' }, forWordlist)
+    lookupText: async (textSelector) => {
+      return moduleInstance.lexicalQuery({ store, textSelector, source: 'lookup' })
+    },
+
+    /**
+     * Starts a lexical query from a lookup component.
```
balmas, line 445: update comment ("lookup" should be "wordlist")

irina060981: fixed comments

irina060981:
>this looks okay to me, although I think we still have more to do in LexicalQuery to suppress the UI Events when the lookup is for the wordlist.

I think that it is worth to make a refactoring - to create a configurable lexicalQuery (inside getData), but it is titghtly used with different controller through events - so we should check all the event map.
And may be we shouldn't populate full defs, word usage examples, lemmaTranslations for flashcard download.
But otherwise may be later we would use relooked up process for wordItems for some other purposes where all these information would be useful.

ok - will merge it for now

Issue #276, opened 2020-04-28
Questions on enable input in betacode


monzug:
1) while testing issue #184, I noticed that the pop up in the screenshot had the word Mou=sa for Μοῦσα.
if I enter Mo=usa, it shows Μοῦσα as it is suggested in the sample chars.
if I enter Mou=sa, it does not change into Μοῦσα but the output is the same in the popup.
question: should we check for error or is it ok as is?

2) also z = ζ , shouldn't it be z => ζ  ?

balmas:
I'll check.

balmas:
via this issue should also address this:

>q/esqai for θέσθαι -> it worked!!!! I did try few Greek words and it works fine. for πρὸς, shall we add that p is r? or is it redundant?

Yes I think we should add that. All chars that are converted should be listed.

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/issues/104#issuecomment-616559299_

balmas:
re mou=sa, it seems that the code we based this on has reversed the normal order of the betacode accents. The standard is to have the accent marks after the letter and not before.  We need to fix this for all of the letters.

balmas:
we also need to fix use of ! it should be |

balmas:
Ah, I see why they switched the diacritics to before the vowel -- because the javascript to transform the text upon input is much more complex to write if you do it after. I'm going to leave it before.

I have made other changes however, to bring the implementation as close as possible to standard betacode. We should use | instead ! for PROSGEGRAMMENI (unicode 1fbe) and we should convert / into TONOS instead of OXIA (See also note in #239 )

I have also added support for _ for macron and ^ for vrachy and a few other combinations of diacritics that were missing. And I removed support for some non-standard transforms.

balmas:
Alpheios Components 3.3.0-qa.20200501441

monzug:
Looks good to me

PR #290, opened 2020-04-30
I276

balmas:
I have made a few corrections to the Betacode transformation to try to bring it more inline with standard usage (the library we based the implementation on deviated from the standard in a variety of ways -- the only one I didn't correct was the placement of the diacritics as noted in #276)

I also added all of the supported letters (lower-case only) in the hints and switched to using a grid to display them.

I added unit tests for all characters extracted from our standard beta-unicode transformation table. (at https://github.com/alpheios-project/xml_ctl_files/blob/master/xslt/trunk/beta-uni-util.xsl)

@irina060981 I need a little help with the reformatting of the height and width of the lookup box now. It isn't quite right in all 3 ui sizes and also needs some work on mobile. Can you help me with that before we merge this? Thank you!


irina060981:
I have updated styles for lookup box
**large**
image

**medium**
image

**small**
image

balmas:
thank you @irina060981 !

Issue #295, opened 2020-05-01
popup not displaying


balmas:
for some reason in the latest qa build (Alpheios Components 3.3.0-qa.20200501441)
the popup is not displaying at all. No error in the console, it just never gets displayed. it's size is set to 0 and display none.

@kirlat could this be related to any of the changes for resizing?

irina060981:
@balmas and @kirlat , I am really sorry - it is a silly misspell in my code, I have fixed it!

irina060981:
merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200504341

monzug:
no problems with pop-ups in build  Alpheios Reading Tools 3.1.0 build qa.20200507338
Alpheios Components 3.3.0-qa.20200507327

PR #297, opened 2020-05-04
Fix bug in lexis for LexicalQuery

irina060981:
For the issue - https://github.com/alpheios-project/alpheios-core/issues/295

@balmas and @kirlat, It is because of my bug from this PR - https://github.com/alpheios-project/alpheios-core/pull/287

As it is a really tiny bug - would merge it without a review


balmas:
thank you @irina060981 !

Issue #258, opened 2020-04-21
disable treebank functionality after X failed requests


balmas:
Per discussion with @kirlat in Slack:

When a page is configured with treebank references, the workflow allows for a period of retries while Arethusa loads. This can occur with each lexical query and if the page is misconfigured or the server hosting Arethusa is not responding, users will experience delays on each request. 

Given that the treebank application and data gets loaded in the browser, it's not likely to succeed on subsequent requests if the first fails, and vice-versa. i.e. if Arethusa loads successfully for a document, it's likely that all requests for data from that document will succeed quickly and if it fails, it's likely that none will succeed.

So we should mitigate the negative user experience for a misconfigured page or non-responsive server by adding a state flag which disables the treebank functionality until the page is reloaded.

So, in the `Lexis` module, in `refreshUntilLoaded` , if the refresh still fails after the configured number of retries is reached, we should set a state variable which disables all further treebank requests for the page, and issue an event that the ui controller can respond to to notify the user that the treebank functionality is disabled.  The notfication text would be something like:

"Some of the resources referenced by this page for refining lookup results are not responding. Results may be less precise than expected."

balmas:
@kirlat I forgot this is a priority for the release too.

kirlat:
Changes are in master now.

balmas:
Alpheios Components 3.3.0-qa.20200507327

need to create a test page for this still

balmas:
this can be tested via https://alpheios-misc-dev.s3.us-east-2.amazonaws.com/treebank-test-page/test.html

balmas:
There's something not quite right about this. Testing with the Odyssey at https://texts-test.alpheios.net/text/urn:cts:greekLit:tlg0012.tlg002.alpheios-text-grc1/passage/1.1-1.30 the first few requests work fine, but then suddenly we stop getting treebank data even when we should.

Needs further investigation.

balmas:
There might be a few things going wrong, but one scenario is related to selection and it exposes a flaw in the logic for how we determine when not to try the treebank requests anymore:

[...]

PR #299, opened 2020-05-04
Do not retry reqeusts to the failed URL

kirlat:
Do not try to send request to treebank again if previous one to the same URL failed. See #258 for more details.


packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -190,7 +210,7 @@ export default class Lexis extends Module {
     this._treebankDataItem = treebankDataItem
     let annotatedHomonyms
 
-    if (this._treebankReady && this._treebankDataItem) {
+    if (this._treebankReady && this._treebankDataItem && !this.isFailedProvider(this._treebankDataItem)) {
```
balmas, line 213: under what circumstances would _treebankReady be true AND isFailedProvided also be true?

kirlat: I think you're right, they are exact opposites now. `_treebankReady` has an additional functionality before that is now not used now. I think there is no reason for `_treebankReady` any more. I will update the code to reflect that.

Issue #288, opened 2020-04-30
Questions/suggestion on mousemove event


monzug:
related to issue #169
1) if I check mouse over (or move) for Chinese option, the login to save your word message is not longer generated in the pop-up. in Google docs, where mousemove is the default, the message is still generated. why such a difference?

2) to simplify the Mouse Move (or over) event, could we have an advanced section where we can configure the delay, the accuracy, Google docs container id?

3) I did also notice that the two digit words are not always selected on mousemove: sometimes it just ignored it, sometimes it does highlighte the line and generate a wrong pop-up, sometimes it does process the two digit word.
see examples below in Google doc

a) Ὅσοις τῶν ἀνθρώπων ἐκ τῆς αὑτῶν2 ὁρμωμένοις χώρας ὑπερόριοί τε ἀγῶνες καὶ κίνδυνοι συμβαίνουσιν, ἄν τι σφάλμα γένηται
- mousemove on τι  does not do anything. if I enter it in Lookup it says it's a pronoun. to generate the pop-up I need to move the mouse almost over the next word.
- mousemove on ἐκ  or  ἄν   works fine (though, I need to position the mouse at the very end of each word)

b) μάλιστα δὲ πρὸς τὴν φύσιν· ἔστι γὰρ οἷον ἀρχὴ
τῶν ζῴων. Ἐπιζητοῦμεν δὲ θεωρῆσαι καὶ γνῶναι τήν τε 
φύσιν αὐτῆς καὶ τὴν οὐσίαν, εἶθ’ ὅσα συμβέβηκε περὶ αὐτήν
- mousemove on  τε  it does highlight the line and generate a pop-up for the previous word. If I move the mouse at the very end of the two digit word, it does not do anything.
see screenshot
![te-in-popup](https://user-images.githubusercontent.com/41396793/80712292-c82cfe80-8abf-11ea-859a-7073998d010b.JPG)

balmas:
>f I check mouse over (or move) for Chinese option, the login to save your word message is not longer generated in the pop-up. in Google docs, where mousemove is the default, the message is still generated. why such a difference?

That's a good question. @irina060981 can you investigate?

balmas:
>to simplify the Mouse Move (or over) event, could we have an advanced section where we can configure the delay, the accuracy, Google docs container id?

Agree this is a good idea.

balmas:
[...]

PR #303, opened 2020-05-06
I288 mousemove fixes

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/288

>to simplify the Mouse Move (or over) event, could we have an advanced section where we can configure the delay, the accuracy, Google docs container id?

I have re-defined the following options as UI options - `mouseMoveDelay`, `mouseMoveAccuracy`, `enableMouseMoveLimitedByIdCheck`, `mouseMoveLimitedById`

because 
 - they are UI options by sense
 - it would make tabs more clear

now tabs look like the following:

image

image


packages/components/src/lib/selection/media/html-selector.js
```diff
@@ -256,8 +257,8 @@ export default class HTMLSelector extends MediaSelector {
     }
     // clean string:
     //   convert punctuation to spaces
-    anchorText = anchorText.replace(new RegExp('[' + textSelector.model.getPunctuation() + ']', 'g'), ' ')
 
+    anchorText = anchorText.replace(new RegExp('[' + textSelector.model.getPunctuation() + ']+', 'g'), ' ')
```
irina060981, line 261: this is the fix for 
> I did also notice that the two digit words are not always selected on mousemove: sometimes it just ignored it, sometimes it does highlighte the line and generate a wrong pop-up, sometimes it does process the two digit word.
see examples below in Google doc

the problem here that, Google docs converts spaces to special characters, 
and they are converted with this Regexp to several spaces, that could braek words definition

irina060981:
> nice catch on the punctuation regex.
> 
> It's okay to merge but I'm not sure about the options change. I was thinking maybe we would have a new "Advanced" options tab, which we could put things like the mousemove and google docs settings that we don't expect people to change, but want to make available in case we need to tell them to.

ok, then I would merge it and we could continue discussion about options tabs later and change it if you think it should be on Advanced.

From my point of view Advanced tab is something like Other tab, noone could be sure - what should be placed there :) that's why I decided to put these settings to UI tab.

balmas:
>From my point of view Advanced tab is something like Other tab, noone could be sure - what should be placed there :) that's why I decided to put these settings to UI tab.

Ok let's see what @monzug thinks

monzug:
I like the Advanced tab where we can put more stuff too (like Log level). otherwise it should go in UI

Issue #209, opened 2020-03-24
don't show Greek pronoun table if no inflection class matches


balmas:
Related to #208 

If we can't determine the pronoun class for a Greek pronoun, we shouldn't show any table for it

irina060981:
Merged and closed

PR #308, opened 2020-05-07
Fix for pronouns without grmClass

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/209

small fix for creating InlfectionData and the test for it

I made changes only in `inflection-tables` packages, all other changes are from `es-lint`


packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -369,28 +369,34 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
     // if we still don't have a match, we can do a last ditch check without
     // any diacritics at all in those languages that support it
     if (!found) {
-      let lastAlt = []
+      const lastAlt = []
```
balmas, line 372: I can never remember, but don't we need to define arrays with let if new values are going to be pushed to them?

balmas: (if so, this is my mistake for not including the necessary eslint here I know...)

irina060981: yes - here should be special es-lint commands
I didn't touch this package during development - that's why I decided not to fix it bfore review - I would add this es-lint rule

balmas: thank you!

PR #314, opened 2020-05-12
Small fix for selenium testing

irina060981:
Added id for lookup field via `input-autocomplete` component


irina060981:
@balmas and @kirlat , it is really a small fix - I will merge it without approval

Issue #291, opened 2020-05-01
suppress events for lookup for wordlist export


balmas:
we should probably suppress UI Events when we are doing a lookup for the wordlist export

balmas:
@kirlat can you take a look at this one? I think it probably needs to be done before the release because I'm a little worried about the side effects of the wordlist lookups. Thanks!

kirlat:
Sure, will work on it

kirlat:
It has been merged to master now

balmas:
Alpheios Components 3.3.0-qa.20200513442

balmas:
@monzug a test scenario for this is the following:

(1) login 
(2) lookup a bunch of words so that they get added to your wordlist 
(3) logout 
(4) clear your browser cache
(5) login again 
(6) do a lookup of some other word (i.e. not one on your wordlist, maybe in a different language)
(7) export some items from your wordlist in flashcard format
(8) make sure nothing weird is happening in the UI as a result of your export -- i.e. state should be the same as after the lookup in step 6

Let me know if you have any questions about this. Thanks!

monzug:
well there are few things that do not look ok.

1) Latin library in Chrome: the checkbox for flashcards or with filtering cannot be checked. however it works in Loeb Classics.

2) the downloaded file with flashcards option is empty. There is only word definition in it. I tried with latin (more than 25 words), then I tried with Persian (3 words). 

3) when I downloaded the Persian file with the flashcard option, for same reason it did a lookup of the last Greek word from the above word list. see screenshot

![weird-scenario](https://user-images.githubusercontent.com/41396793/82055288-42e44500-968e-11ea-9605-d07fd33a8709.JPG)

monzug:
2) could it be because I haven't updated the env-webext.js yet?  I read your reccomendation in #200 after testing this one

monzug:
as per 2) I was able to download the file with the flashcard option. however if I chose to limit to 1 the most  often list (of two words), the downloaded file listed both words.

as per 1) - I think it's the first language (it can be Greek or Latin) on the wordlist that does not work.

irina060981:
@kirlat and @balmas , I tested a simple scenario - when we have data only in remote source and don't have data in IndexedDB - and homonyms are not updated without `HOMONYM_READY`
and finally - we have an empty wordlist downloaded

This happens because we send textSelector to lexis module (without any link to wordItem),
and then wordListController updates worditems from the same events:

```
[...]

PR #321, opened 2020-05-12
wordlist lookup

kirlat:
Here are the changes to suppress UI messages for wordlist queries. Tested it for different words and it seems to be working without any issue.

Also added a clause to adapters' code to prevent it from throwing an error when data is incomplete.

See #291 for related information.


packages/components/src/lib/queries/lexical-query.js
```diff
@@ -274,19 +299,23 @@ export default class LexicalQuery extends Query {
...
-          opts: lexiconFullOpts,
-          homonym: this.homonym,
-          callBackEvtSuccess: LexicalQuery.evt.FULL_DEFS_READY,
-          callBackEvtFailed: LexicalQuery.evt.FULL_DEFS_NOT_FOUND
-        }
-      })
+      let adapterLexiconResFull = {}
+      // Do not retrieve full definition for wordlist requests
+      if (this._source !== LexicalQuery.sources.WORDLIST) {
+        adapterLexiconResFull = yield ClientAdapters.lexicon.alpheios({
```
irina060981, line 305: @kirlat , why did you decide to turn off full homonyms for wordlist?
We save them to IndexDB source and show when a user clicks on a word in wordlist.

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -166,30 +167,45 @@ export default class LexicalQuery extends Query {
         if (this._annotatedHomonyms && this._annotatedHomonyms.hasHomonyms) {
           this.homonym = Homonym.disambiguate(this.homonym, this._annotatedHomonyms.homonyms)
         }
-        LexicalQuery.evt.MORPH_DATA_READY.pub()
+        // Suppress events that will trigger UI messages if source is wordlist
+        if (this._source !== LexicalQuery.sources.WORDLIST) {
+          LexicalQuery.evt.MORPH_DATA_READY.pub()
+        }
```
irina060981, line 173: The following events - HOMONYM_READY, SHORT_DEFS_READY, LEMMA_TRANSL_READY
executes update provider service, and we save provider information to IndexDB
But I am not sure , that we retrieve that information correctly

packages/components/src/lib/queries/lexical-query.js
irina060981: Oh, I believe I know, why. Because we use only short definitions. 
I didn't turn off it myself in my PR because I believe that it could be usefull to update data and for full homonyms too, as this process is async and won't make delays for quizlet download.

kirlat: So if we do not need full definitions for word list right now maybe we can disable its retrieval for now to make things faster, but we could enable it later. @balmas, @irina060981, what do you think?

packages/components/src/lib/queries/lexical-query.js
kirlat: This is a good catch! If we need to update providers during wordlist requests then we should probably take provider retrieval code out of a UI controller (which would be a good thing from an architectural standpoint, on my opinion). 

Right now callbacks of those events in a UI controller update both the UI and retrieve provider data. We could change it so that a UI controller will be responsible for updating UI only and register another subscriber to those events in probably LexisCS that will be responsible for getting provider data.

The question is how critical is that. I'm afraid to do such changes before a release, it would affect too many parts. If we need it, maybe we can create an issue and do it after the release?  @balmas, what do you think?

packages/components/src/lib/queries/lexical-query.js
balmas: yes, I think we can disable it for now.

packages/components/src/lib/queries/lexical-query.js
balmas: I agree with your suggestions on refactoring and agree it can wait. The provider information is not really necessary for the wordlist download and in any event, I think what the UI Controller code does is dedupe the provider information that is already there, so at the worst we end up with repetitive information.  Please do enter an issue for addressing this as you suggest in a future release though. Thank you both!

packages/components/src/lib/queries/lexical-query.js
irina060981: By the way - it won't make things faster, as full definitions are retrieved completely async, and it doesn't influence on a delay before download starts

balmas: not for the UI but it could add load to the server for unnecessary requests

PR #335, opened 2020-05-14
Reworking of tagged commit script

kirlat:
This is an update to use the latest features of `node-build`. It allows `node-build` to be hoisted so that all packages will use the same version of it. I've made necessary changes to the root `package.json` in order for it to work. For alpheios-project/node-build#32.

Will update `node-build` tags after it is merged to `v3`.


packages/client-adapters/package.json
```diff
@@ -36,7 +36,7 @@
     "alpheios-data-models": "^3.3.0",
     "alpheios-fixtures": "github:alpheios-project/fixtures",
     "alpheios-messaging": "github:alpheios-project/alpheios-messaging",
-    "alpheios-node-build": "github:alpheios-project/node-build#v3",
+    "alpheios-node-build": "github:alpheios-project/node-build#i32-script-optimization",
```
balmas, line 39: needs to be fixed before merge

tagged-commit.mjs
```diff
@@ -1,50 +1,63 @@
...
-try {
-  output = execSync(`git tag ${buildInfo.name}`, { encoding: 'utf8' })
-} catch (error) {
-  console.error('Tagging failed:', error)
-  process.exit(5)
-}
-console.log(`Commit has been completed`)
+  console.log(`Tagging with ${buildInfo.name}`)
+  try {
+    //output = execSync(`git tag ${buildInfo.name}`, { encoding: 'utf8' })
```
balmas, line 57: uncomment?

tagged-commit.mjs
```diff
@@ -1,50 +1,63 @@
...
-}
-try {
-  output = execFileSync('git', ['commit', '-m', `Build ${buildInfo.name}`], { encoding: 'utf8' })
-} catch (error) {
-  console.error('Commit process failed:', error)
-  process.exit(4)
-}
+  console.log('Committing components/dist')
+  try {
+    //output = execFileSync('git', ['add', 'lerna.json', 'packages/*/package*.json', 'packages/components/dist'], { encoding: 'utf8' })
```
balmas, line 43: uncomment?

tagged-commit.mjs
```diff
@@ -1,50 +1,63 @@
...
+  } catch (error) {
+    console.error('Commit process failed:', error)
+    process.exit(3)
+  }
+  try {
+    //output = execFileSync('git', ['commit', '-m', `Build ${buildInfo.name}`], { encoding: 'utf8' })
```
balmas, line 49: uncomment?

tagged-commit.mjs
```diff
@@ -1,50 +1,63 @@
...
+      module: 'webpack',
+      mode: 'all',
+      preset: 'vue',
+      codeAnalysis: false,
+      outputLevel: Builder.outputLevels.MIN,
+      buildTime: buildDT
+    })
+    await builder.importConfig('config-tagged-commit.mjs', 'packages/components/build')
+    await builder.runModules()
+    // output = execSync(`cd packages/components && node --experimental-modules ./node_modules/alpheios-node-build/dist/build.mjs --module=webpack --mode=all --preset=vue --externalConfig=config.mjs --buildTime=${buildDT}`, { encoding: 'utf8' })
```
balmas, line 34: remove commented code?

tagged-commit.mjs
```diff
@@ -1,50 +1,63 @@
...
-} catch (e) {
-  console.error('Cannot execute npm version:', e)
-  process.exit(1)
-}
+  const baseVersion = pkg.version.split('-')[0]
+  const version = `${baseVersion}-${buildInfo.name}`
+  console.log(`Setting a package version to ${version}`)
+  let output
+  try {
+    //const output = execSync(`npx lerna version ${version} --no-git-tag-version --no-push --yes`, { encoding: 'utf8' })
```
balmas, line 16: uncomment?

packages/components/build/config-tagged-commit.mjs
```diff
@@ -0,0 +1,91 @@
+/*
```
balmas, line 1: how does this config differ from the one used to build components normally? I.e. the config.mjs here. Is there a reason we can't just reuse that config here?

packages/client-adapters/package.json
kirlat: Yes, will do

tagged-commit.mjs
kirlat: Sorry, will fix it

tagged-commit.mjs
kirlat: Will fix

tagged-commit.mjs
kirlat: will remove a comment

tagged-commit.mjs
kirlat: will do

tagged-commit.mjs
kirlat: will update

packages/components/build/config-tagged-commit.mjs
kirlat: It has different paths. In order to avoid duplicating a config file we can modify `node-build` so that it, instead of a config file name, will be able to to take a config object. Then we can load a config file in a tagged commit, modify paths, and provide an updated config file to the `Builder` class. @balmas, what do you think about a solution like that?

That would open some other interesting possibilities because we could have many more build parameters defined in a config object (file) than we do now. It will make things much more flexible for future extension.

balmas: I think that's okay, but let's postpone it until after the release since it isn't essential.

kirlat: OK. The reason I decided to have two copy of config files because `alpheios-core` is the only place where we need that. It is not required neither for `embed-lib`, nor for `webextension`.

I will enter an issue for that.

balmas: thanks!

irina060981:
@kirlat , sometimes I face with problems with updating packages for testing, that's why
I would like to ask - could you explain the following:
  - why do we have this outdated libs inside package.json packages, for example
    ```
    "alpheios-client-adapters": "^3.3.0",
    "alpheios-data-models": "^3.3.0",
    ```
    - why do we need to have `alpheios-fixtures` as a devDependency inside each package and not in the root folder

    - do we use `alpheios-fixture` from the local package node-modules inside tests?

    - in package.json of components we have these settings:
    ```
    "moduleNameMapper": {
      "^@vue-runtime$": "vue/dist/vue.runtime.common.js",
      "^@[/](.+)": "<rootDir>/src/$1",
      "^@tests[/](.+)": "<rootDir>/tests/$1",
      "alpheios-data-models": "<rootDir>/node_modules/alpheios-data-models/dist/alpheios-data-models.js",
      "alpheios-messaging": "<rootDir>../../node_modules/alpheios-messaging/dist/dev/alpheios-messaging.js",
      "alpheios-morph-client": "<rootDir>/node_modules/alpheios-morph-client/dist/alpheios-morph-client.js",
      "alpheios-inflection-tables": "<rootDir>/node_modules/alpheios-inflection-tables/dist/alpheios-inflection-tables.js"
    },
```
   these way what packages are used inside tests?

Would be glad to have your thoughts about this :)

kirlat:
> why do we have this outdated libs inside package.json packages, for example
> "alpheios-client-adapters": "^3.3.0",
> "alpheios-data-models": "^3.3.0",

This is the current version of those packages as listed in their `package.json` files: https://github.com/alpheios-project/alpheios-core/blob/master/packages/client-adapters/package.json#L3. `alpheios-core` is managed by Lerna, and we've chosen all packages to to have the same version. Once we use Lerna version tool (mostly via npm scripts in `alpheios-core/package.json`), it will increase version of all packages synchronously. With our current workflow model, we would not do any version increases until the code is merged into a `production` branch. Because this did not happen yet, we will still have `3.3.0` until then.

kirlat:
> why do we need to have alpheios-fixtures as a devDependency inside each package and not in the root folder

The principle of Lerna controlled directory is that each subpackage lists its dependencies as if they were completely independent. Lerna takes dependencies from `package.json` of each subpackage, builds a list of all dependencies, and installs them all. They can be installed into the `node_modules` of `alpheios-core` or into the `node_modules` of individual subpackages, depending on Lerna's `host` and `nohoist` options.

Idea behind that is that even though all packages are in one directory, each package is still independent. Each package can, for, example. be published to `npm` on its own. So those users who got it from npm should be able to use it independently of other packages. That requires dependencies to be listed in `package.json` of each individual subpackage, on on `package.json` of Lerna root. 

To put it simply, the principle is: Lerna knows about subpackages it owns, but subpackages should know nothing about Lerna.

kirlat:
> do we use alpheios-fixture from the local package node-modules inside tests?

No, I think we're not. At least that's how it is on my machine. We're using it from `alpheios-core/node_modules`. This works because the package resolution algorithm searches for package in `node_modules` of the current directory and, if not found, in `node_modules` of parent directory and so on, until it finds it in `alpheios-core`.

kirlat:
>   in package.json of components we have these settings:
> ```
>     "moduleNameMapper": {
>       "^@vue-runtime$": "vue/dist/vue.runtime.common.js",
>       "^@[/](.+)": "<rootDir>/src/$1",
>       "^@tests[/](.+)": "<rootDir>/tests/$1",
>       "alpheios-data-models": "<rootDir>/node_modules/alpheios-data-models/dist/alpheios-data-models.js",
>       "alpheios-messaging": "<rootDir>../../node_modules/alpheios-messaging/dist/dev/alpheios-messaging.js",
>       "alpheios-morph-client": "<rootDir>/node_modules/alpheios-morph-client/dist/alpheios-morph-client.js",
>       "alpheios-inflection-tables": "<rootDir>/node_modules/alpheios-inflection-tables/dist/alpheios-inflection-tables.js"
>     },
> ```
> 
> then what packages are used inside tests?

Let's say we're running tests inside `components`. Then the `<rootDir>` would be `alpheios-core/packages/components`. We will use `alpheios-data-models` from `alpheios-core/packages/components/node_modules`. During bootstrap, Lerna creates symlinks to all subpackages within their local directories, so we have to use a local directory there.

`alpheios-messaging`, however, is an external package. Because we're using a `hoist` option for Lerna bootstrap (except for `coveralls` and `eslint-config-standard`), it will be installed to `alpheios-core/node_modules`. That's why we use `../..` before its path, to point to a folder two levels higher.

kirlat:
Please let me know if anything is still unclear or if you have more questions. I would be glad to help! Answering questions helps me to formalize it in my head better 🙂.

irina060981:
@kirlat , thank you for your detailed answers! It was very helpful to understand how it works.

Than I have one more question :
I have updated alpheios-fixtures with new fixture and it has an updated version in master branch
I work with tests inside `inflection-tables` package
    - where could I update this lib to update tests - inside `inflection-tables` package?
    - where would travis check dependency - what package.lock.json should I remove?

Now I update and remove from all packages that has this dependency to avoid problems
But may be there is some more effective way to do?

kirlat:
> - where could I update this lib to update tests - inside `inflection-tables` package?

Yes, I think this is the place. If fixtures are merged to master, then you can probably do `npm run bootstrap` from the root of  `alpheios-core` and Lerna should install an updated `alpheios-fixtures`.

If you want to install fixtures of a specific tagged version or branch you can specify this branch or version inside dependencies of `inflection-tables`. In that case, I believe, a bootstrap command will install that specific version into `alpheios-core/packages/inflection-tables/node_modules` and the "regular" version from master will be installed into  `alpheios-core/node_modules`. As a result, `inflection-tables` will use that specific version and the rest of packages will use a version from master.

If there is a need to force specific version on all packages, you can edit `package.json` files of all subpackages and specify that version there. Or you can change it in one package and run `list-mismatching-deps` script that will show what packages list a different version of it. Then you can fix it manually. There is a `fix-mismatching-deps` script but I think it will not work reliably in this case because it cannot know if one branch is "higher" than the other and what branch is the newest version.

irina060981:
Great! Things become much clearer! Thank you :)

balmas:
> > * where could I update this lib to update tests - inside `inflection-tables` package?
> 
> Yes, I think this is the place. If fixtures are merged to master, then you can probably do `npm run bootstrap` from the root of `alpheios-core` and Lerna should install an updated `alpheios-fixtures`.
> 
> If you want to install fixtures of a specific tagged version or branch you can specify this branch or version inside dependencies of `inflection-tables`. In that case, I believe, a bootstrap command will install that specific version into `alpheios-core/packages/inflection-tables/node_modules` and the "regular" version from master will be installed into `alpheios-core/node_modules`. As a result, `inflection-tables` will use that specific version and the rest of packages will use a version from master.
> 
> If there is a need to force specific version on all packages, you can edit `package.json` files of all subpackages and specify that version there. Or you can change it in one package and run `list-mismatching-deps` script that will show what packages list a different version of it. Then you can fix it manually. There is a `fix-mismatching-deps` script but I think it will not work reliably in this case because it cannot know if one branch is "higher" than the other and what branch is the newest version.

I have found that things work a little inconsistently with (at least) alpheios-fixtures.  If I want to use a branch version for testing of inflection-tables, sometimes it works to update the inflection-tables package.json dependency to use the branch and run `npm bootstrap` and sometimes it doesn't.  I haven't figure out why it doesn't always work, usually when that happens I just merge the fixtures to master and use that version. 

But another thing I've noticed is that for the hoisted packages, i.e. those in alpheios-core/node_modules, `npm clean` and  `npm bootstrap` do not seem to touch them. If they've been updated I have to delete them manually and then rerun bootsrap.

Issue #339, opened 2020-05-14
can we add the lang attribute to the lookup input box?


balmas:
@monzug notes in #320 that the lookup box uses a different font for syriac and arabic than we do on the popup. This is because we use the lang attribute to trigger better fonts for those languages in our css. Is it possible to apply the same to the lookup input box?

irina060981:
Fixed 

![image](https://user-images.githubusercontent.com/12377640/82032778-186ea800-96df-11ea-82c1-10c321724c0b.png)

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200522457

monzug:
looks ok to me.

Issue #338, opened 2020-05-14
language specific font settings being overruled in Loeb Classics


balmas:
![a-syriac-name](https://user-images.githubusercontent.com/41396793/81941063-55934700-95c6-11ea-9ea5-e39b94a531a5.JPG)

_Originally posted by @monzug in https://github.com/alpheios-project/alpheios-core/issues/320#issuecomment-628639914_

irina060981:
Fixed 

![image](https://user-images.githubusercontent.com/12377640/82032778-186ea800-96df-11ea-82c1-10c321724c0b.png)

![image](https://user-images.githubusercontent.com/12377640/82032855-320fef80-96df-11ea-996d-b2b6901f923b.png)

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200522457

monzug:
I tested few words in Loeb classics and it looks ok.

Issue #334, opened 2020-05-14
Add extra space between beta chars and the separation line below it


balmas:
I wouldn't mind to add extra space between the sample characters and the separation line. @balmas?

_Originally posted by @monzug in https://github.com/alpheios-project/alpheios-core/issues/329#issuecomment-628598818_

irina060981:
Fixed 
![image](https://user-images.githubusercontent.com/12377640/82034067-042baa80-96e1-11ea-94d8-6074c98bdf7c.png)

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200522457

monzug:
I don't think this fix is the build  Alpheios Reading Tools 3.1.0 build qa.20200522460
Alpheios Components 3.3.0-qa.20200522457

irina060981:
it is trange - but it is missed in the most last version in repository

irina060981:
and I could see that the height of the lookup box is bigger than in my change.
Will do this changes again to the code

irina060981:
also I could see that border styles for Loeb missed again - will fix them too
![image](https://user-images.githubusercontent.com/12377640/82860061-dace1280-9f5b-11ea-9835-08a94be5cfd0.png)

irina060981:
merged and closed

balmas:
Alpheios Reading Tools 3.1.0 build qa.20200527411
Alpheios Components 3.3.0-qa.20200527404

monzug:
space and border style are fixed. Thanks

PR #342, opened 2020-05-15
Syriac fonts fix

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/339
For the issue https://github.com/alpheios-project/alpheios-core/issues/338

set Syriac  lang for 
  - lookup input
  - infl-attribute for source

fix Loeb style for Headings with syriac
`font-variant-ligatures: normal;`

Also added fix for lookup height

image

image

For the issue https://github.com/alpheios-project/alpheios-core/issues/334
added padding

image

For the issue https://github.com/alpheios-project/alpheios-core/issues/288
fixed Mouseover and fieldset styles (Loeb)

image


packages/components/src/styles/_content.scss
```diff
@@ -18,11 +18,13 @@
   [lang='ara'] {
     // `Important` below is required to override aggressive styling on pages
     font-family: var(--alpheios-arabic-font-face) !important;
+    font-variant-ligatures: normal;
   }
 
   [lang='syr'] {
     // `Important` below is required to override aggressive styling on pages
     font-family: var(--alpheios-syriac-font-face) !important;
+    font-variant-ligatures: normal;
```
balmas, line 27: does this need an !important flag? And would it be a good idea to add it for everything not just the syriac font?

irina060981:
```
font-variant-ligatures: normal;
```

In this case we don't need an important flag, because it doesn't have important on the site.
I believe that it seems something like use or not use other characters for headings.
Loeb has this styles
```
h1, h2, h3, h4, .hero-content h1 {
    font-family: "zephtextregular";
    font-variant-ligatures: none;
}
```

But I am not sure with all languages and this property - I haven't any experience with it
That's why I added only to `syr` and `ara` as they have specific fonts

I think if we want to add such a property to other languages - we first need to collect words for testing those language
@balmas , what do you think?

balmas: Let's enter an issue for it and see if it comes up again before doing any more

PR #356, opened 2020-05-19
I347 syriac

balmas:
updates the punctuation list for Syriac


packages/data-models/src/syriac_language_model.js
```diff
@@ -85,11 +85,12 @@ for the current node
 
   /**
    * Get a list of valid puncutation for this language
+   * Taken from  the list at https://en.wikipedia.org/wiki/Syriac_(Unicode_block)
    *
    * @returns {string} a string containing valid puncutation symbols
    */
   static getPunctuation () {
-    return "፡፨።፣፤፥፦፧፠,;:!?'\"(){}\\[\\]<>/\\\u00A0\u2010\u2011\u2012\u2013\u2014\u2015\u2018\u2019\u201C\u201D\u0387\u00B7\n\r\u200C\u200D"
+    return "\u0700\u0701\u0702\u0703\u0704\u0705\u0706\u0707\u0708\u0709\u070A\u070B\u070C\u070D\u070F\.,;:!?'\"(){}\\[\\]<>/\\\u00A0\u2010\u2011\u2012\u2013\u2014\u2015\u2018\u2019\u201C\u201D\n\r\u200C\u200D"
```
kirlat, line 93: Not sure if we need a backslash before a dot: `\.`. Dot literals are not escaped in a character class, so probably the combination would be interpreted as `\` and `.`, both literal characters. We already have `\` in a set later and it will probably change nothing, but just wanted to check with you.

balmas: thanks, will fix that.

PR #358, opened 2020-05-19
I4 alpheios trees

balmas:
see https://github.com/alpheios-project/alpheios-trees/issues/4#issuecomment-631007920


packages/client-adapters/src/adapters/arethusa/adapter.js
```diff
@@ -110,6 +110,14 @@ class ArethusaTreebankAdapter extends BaseAdapter {
...
         const transformAdapter = new AlpheiosLexiconTransformer(this, mapper, 'arethusa')
         const homonym = transformAdapter.transformData(tbRes, word)
+        // handle verb participles in a way consistent with the morpheus parser
+        // which reports the pofs of the lemma as verb and pofs of the inflection as verb participle
+        if (homonym.lexemes.length == 1 &&
+           homonym.lexemes[0].lemma.features[Feature.types.part].value === Constants.POFS_VERB &&
+           homonym.lexemes[0].inflections.length == 1 &&
+           homonym.lexemes[0].inflections[0][Feature.types.mood].value === Constants.MOOD_PARTICIPLE) {
+             homonym.lexemes[0].inflections[0].addFeature(new Feature(Feature.types.part,Constants.POFS_VERB_PARTICIPLE,languageModel.languageID))
+        }
```
kirlat, line 120: Have a question about this: why we check and correct only the first inflection of the first lemma? Is verb participle in other positions not possible?

balmas: because with treebank data, we will only ever get 1 lemma and 1 inflection

Issue #357, opened 2020-05-19
fix formatting of wordlist


balmas:
A couple of points, possibly mentioned already elsewhere:

* I think need to prevent the date from wrapping to a new line
* the sort arrows need to be centered over the data column

![Screenshot from 2020-05-19 13-58-43](https://user-images.githubusercontent.com/644870/82361533-f889fd80-99d8-11ea-8667-42d357e2f4be.png)

irina060981:
Merged and fixed

balmas:
Alpheios Components 3.3.0-qa.20200522457

monzug:
looks ok now
![ancora-limit-wordlist](https://user-images.githubusercontent.com/41396793/82830057-b81bfb00-9e82-11ea-8216-c8cd06512031.JPG)

PR #363, opened 2020-05-21
Fix bugs with wordlist downloads with short definitions

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/291

**to client adapters**
  - lexicon adapter has `sync` and `async` way of fetching data
  - `async` would be chosen if callbacks are defined
  - otherwise `sync` would be chosen
  
  - also I have updated fixtures usage for lexicon to made more usable for lexicon tests

 
**to components**
  - allow sending HOMONYM_READY for updating homonym with wordlist flag (this attaches homonym to worditem)
  - remove callbacs from lexicon request for wordlist - to activate sync fetch

 - also I have fixed a small bug for wordlistMaxFlashcardExport otherwise wordlist download for flashcard is empty for unlogged users


packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -130,6 +131,68 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
...
+        await this.fetchFullDefsSync(homonym)
+      }
+    }
+  }
+
+  /**
+  * This is a sync method that retrieves short definitions for homonym synchronously
+  * @param {Homonym} homonym - homonym for retrieving definitions
+  */
+  async fetchShortDefsSync (homonym) {
```
kirlat, line 150: It is a little confusing, on my opinion, to have a sync function being declared as async. Maybe it's just me, but it took me a while to grasp it. If I understand correctly, this function retrieves multiple short definitions sequentially. So maybe it's worth to rename it, removing the `sync` part and adding some indication that it retrieves multiple definitions, in contrast with the existing function? What do you think?

packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -130,6 +131,68 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
...
+    } catch (error) {
+      this.addError(this.l10n.messages.LEXICONS_FAILED_CACHED_DATA.get(error.message))
+    }
+  }
+
+  /**
+  * This is a sync method that retrieves full definitions for homonym synchronously
+  * @param {Homonym} homonym - homonym for retrieving definitions
+  */
+  async fetchFullDefsSync (homonym) {
```
kirlat, line 173: Same question about naming as with short definitions

packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -130,6 +131,68 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
...
     }
+
+    if (this.async) {
+      return this.fetchDefsWithCallbacks(homonym, lookupFunction)
+    } else {
+      if (lookupFunction === 'short') {
+        return this.fetchShortDefsSync(homonym)
+      } else if (lookupFunction === 'full') {
+        await this.fetchFullDefsSync(homonym)
+      }
```
kirlat, line 142: Why to we return short defs without waiting for the result to arrive while awaiting on full definitions? Shouldn't it be handle in the same manner?

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -285,15 +283,22 @@ export default class LexicalQuery extends Query {
...
+      let params = {
+        opts: lexiconShortOpts,
+        homonym: this.homonym
+      }
+
+      if (this._source !== LexicalQuery.sources.WORDLIST) {
+        params = Object.assign(params, {
+          callBackEvtSuccess: LexicalQuery.evt.SHORT_DEFS_READY,
+          callBackEvtFailed: LexicalQuery.evt.SHORT_DEFS_NOT_FOUND
+        })
```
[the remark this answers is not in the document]
irina060981: Of course we could, but from my point of view it looks more clearer, and we could easiy see what properties are added

@kirlat , do you see any problems with such syntax?

packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -1654,7 +1654,10 @@ If no URLS are provided, will reset grammar data.
     this.store.commit('app/setQueryStillActive', true)
   }
 
-  onHomonymReady (homonym) {
+  onHomonymReady (homonym, source) {
+    if (this._source === LexicalQuery.sources.WORDLIST) {
+      return
+    }
```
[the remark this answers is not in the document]
irina060981: I think it is a good idea!

packages/client-adapters/src/adapters/lexicons/adapter.js
irina060981: It could return several definitions or could return only one definition, it depends given options.

As for me, async is a special word that shows that this method is working with Promises, and it doesn't tell if actions would be done in parallel or in sequence, it could be understood from the following code.

I have used to type of function names:
 
`fetchDefsWithCallbacks` (here we have both calls for short and full)
`fetchShortDefsSync`
`fetchFullDefsSync`

`updateFullDefsWithCallbacks`
`updateFullDefsSync`

That's why I used this types of addition - `withCallbacks` and `sync`

but of course I am not very good with english, may be there are some other words that show
that the method executes all requests and wait untill all of them would be retrieved

@kirlat and @balmas , do you have better suggestion?

packages/client-adapters/src/adapters/lexicons/adapter.js
irina060981: Described my opinion above

packages/client-adapters/src/adapters/lexicons/adapter.js
irina060981: good catch - i have missed this - fixed

packages/components/src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
irina060981: added

packages/components/src/lib/queries/lexical-query.js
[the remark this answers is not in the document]
kirlat: No, I'm fine with `Object.assign`. On my opinion, both ways are good (we can also use a spread operator as a third way to do it), I think it's a matter of personal preference.

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -218,18 +218,16 @@ export default class LexicalQuery extends Query {
     if (lexiconShortOpts.allow && lexiconShortOpts.allow.length > 0) {
       this.homonym.lexemes.forEach((l) => { l.meaning.clearShortDefs() })
       // Suppress events that will trigger UI messages if source is wordlist
-      if (this._source !== LexicalQuery.sources.WORDLIST) {
-        LexicalQuery.evt.HOMONYM_READY.pub(this.homonym)
-      }
+      LexicalQuery.evt.HOMONYM_READY.pub(this.homonym)
```
balmas, line 221: maybe remove the comment above about suppressing the events since it's not the case anymore

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -218,18 +218,16 @@ export default class LexicalQuery extends Query {
...
     } else {
       // we won't have any remaining valid short definition requests
       // so go ahead and publish the SHORT_DEFS_READY event so that the UI
       // can update itself
       // but issue the HOMONYM_READY event first otherwise we get errors
       // from the wordlist which expects to see the homonym before definitions
       // Suppress events that will trigger UI messages if source is wordlist
```
balmas, line 228: move the comment about suppressing events to below the homonyn_ready event

packages/components/src/lib/queries/lexical-query.js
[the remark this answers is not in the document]
irina060981: ok, than I will leave this code assignment :)

packages/client-adapters/src/adapters/lexicons/adapter.js
irina060981: @balmas , we already have 

```
fetchShortDefs
fetchFullDefs
```

they are used as methods from client-adapters and navigate request through code

so I couldn't rename methods this way

irina060981: @balmas and @kirlat , may be we will have then

**from client-adapters**
`fetchShortDefs`
`fetchFullDefs`

`fetchDefsAsync` - with callbacks
`fetchShortDefsSync` - sync way
`fetchFullDefsSync` - sync way

?

irina060981:
> I'm having a little trouble understanding how this fully solves the problem. the HOMONYM_READY event will be issued before the short defs query completes won't it? on line 221. So when/how will the wordlist get updated with the shortdefs?

We have the following workflow for updating data to wordlist:

  1 we create a lexical query with targetWord and languageId
  2 lexical query executes morph request that is ended with HOMONYM_READY event
  3 WordlistController catches HOMONYM_READY event and updates worditem in the wordlist with this new homonym object
  4 lexical query executes lexicons request for short definitions without callbacks - synchronously
  5 the result of the request updates attached homonym with short definitions
  6 lexical query becomes ended and download process go on to the next step

@balmas , do you see any errors in the current workflow? should I change anything here (it is our standard worflow for wordlist updating, I think)

packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -288,6 +351,37 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
...
+
+      try {
+        if (fullDefData && fullDefData.match(/alph:error|alpheios-lex-error/)) {
+          const error = fullDefData.match(/no entries found/i) ? 'No entries found.' : fullDefData
+          this.addError(this.l10n.messages.LEXICONS_FAILED_CACHED_DATA.get(error))
+        } else {
+          const provider = new ResourceProvider(config.urls.full, config.rights)
+          const def = new Definition(fullDefData, config.langs.target, 'text/plain', request.lexeme.lemma.word)
+          const definition = await ResourceProvider.getProxy(provider, def)
```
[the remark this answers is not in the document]
irina060981: I will check, it is an old code - I didn't touch it  - copied as-is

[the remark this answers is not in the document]
irina060981: as I could see it doesn't return a Promise
```
  static getProxy (provider = null, target = {}) {
    return new Proxy(target, {
      get: function (target, name) {
        return name === 'provider' ? provider : target[name]
      }
    })
  }
```

will remove await from here

packages/components/src/lib/queries/lexical-query.js
irina060981: oh, I have missed it - fixed

packages/components/src/lib/queries/lexical-query.js
irina060981: fixed

irina060981:
Consists fix for the issue #357
from the approve PR https://github.com/alpheios-project/alpheios-core/pull/364

irina060981:
@balmas and @kirlat , I have worked on your review comments.
Should I change anything else - or I can merge it?

balmas:
> > I'm having a little trouble understanding how this fully solves the problem. the HOMONYM_READY event will be issued before the short defs query completes won't it? on line 221. So when/how will the wordlist get updated with the shortdefs?
> 
> We have the following workflow for updating data to wordlist:
> 
> 1 we create a lexical query with targetWord and languageId
> 2 lexical query executes morph request that is ended with HOMONYM_READY event
> 3 WordlistController catches HOMONYM_READY event and updates worditem in the wordlist with this new homonym object
> 4 lexical query executes lexicons request for short definitions without callbacks - synchronously
> 5 the result of the request updates attached homonym with short definitions
> 6 lexical query becomes ended and download process go on to the next step
> 
> @balmas , do you see any errors in the current workflow? should I change anything here (it is our standard worflow for wordlist updating, I think)

Ok thank you this is very helpful.  No, I don't think further changes are needed then if downloading waits for the short defs to complete before moving on to the next word.

PR #374, opened 2020-05-22
Several wordlist bug fixes

irina060981:
For the issue #200 

 - fixed logic for wordItem createdDT and updatedDT
 - fixed clear filters on adding a new word to wordList


balmas:
although actually @irina060981 I have a question -- what will happen with sorting if a word doesn't have an updatedDate?

irina060981:
>although actually @irina060981 I have a question -- what will happen with sorting if a word doesn't have an updatedDate?

@balmas , it is a good question - it could throw a error in our code - will add a fix

irina060981:
tests from travis fails with connecting to remote source with wordlist

irina060981:
>tests from travis fails with connecting to remote source with wordlist
I think I should recheck tomorrow - when there would be less traffic before merge

the problem was the following:
  in test case I created manually wordItem (not through events) and it didn't have a createdDT
  I fixed the case and it starts work well
  
  but also I have added a check for createdDT to remoteDB serialize method (that create json for sending to remote source
  because empty createdDT would fail the put request with 500 Error that doesn't depend on users's action - so there is no reason to cancel put request in such conditions I think

Issue #361, opened 2020-05-20
google-docs mousemove off by default?


balmas:
i was testing use of the 3.0.x release version of Alpheios on google docs today, and it seems that the double-click is working fine again there so they must have made some changes to the way they were intercepting the double-click event.  

i think given this we might need to reverse the change to automatically enable mousemove lookups on google docs -- we could expose this as an option for the future, but I am worried that it is more invasive than it is needed if the double-click handler is working again.

Looking at our usage logs, it appears that the majority of alpheios usage is coming from Google docs pages right now so this change would effect a lot of users

balmas:
@kirlat I'm wondering if you can look at this for me. I've started working on it in the i361-googledocs branch where I've added a setting which controls whether or not we force mousemove on google docs.

I'm still not sure what we should be doing here. I don't know if I'm just going crazy but it seems to me the behavior on Google Docs, at least in Chrome, is inconsistent, but it appears to be working with the released version of Alpheios (with Components 1.4.7) so it's possible that rather than Google docs breaking our code, our code is at fault here.

It's a little more important than I previously realized to get this right, because it seems like a huge percentage of our usage right now is coming from usage on Google docs. And I'm a little worried about forcing it to use mousemove for selection if we don't have to.

What I was seeing today was that the following code is what fails with Double-click on Google Docs:

in `HTMLSelector.doSpaceSeparatedWordSelection` we have this:

```
    const selection = HTMLSelector.getSelection(this.target)

    let anchor = selection.anchorNode // A node where is a beginning of a selection
    if (!anchor) {
      return
    }
```
At least sometimes it seems that the anchorNode is null when you use double-click on Google Docs in Chrome. 

But when I use the 3.0.x release version of Alpheios (with components 1.4.7) it's working at least sometimes.  

Can you investigate -- you should probably start with my i361-googledocs which doesn't force the mouseover behavior on Google Docs so that you can be sure you're exercising the double-click handler.

Thanks!

kirlat:
I will be glad to work on this!

kirlat:
This has been merged to master

balmas:
oh, this should be in the latest QA build

Alpheios Components 3.3.0 build i400.20200529319

monzug:
Mouse movement is not checked by default. switching from double clicking to mouse movement works fine.
Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

PR #375, opened 2020-05-22
Disable Google Docs by default

kirlat:
This change make double click, not mousemove, a default select text event handler on Google Docs. See #361 for details.

I've tried to do as little changes as possible, considering a timeline. I've also removed some extra mousemove registration calls where they were duplicated (it was called twice sometimes). 

As selection workflow became more complex, I am not sure about whether a matching UX is clear enough. For example, if page language is set to Chinese and user is on Google Docs, selection will not work at all. This can be confusing. It was confusing several times for me, a person knowing mechanics better than an Alpheios user.

Another thing in the UX that bothers me a little is that selection now works differently depending on what page language is set (Chinese vs other languages). If page language is set to Chinese, Chinese words are selected well, but words in other languages are not so, and vice versa. I'm wondering if users would realize that it might depend on what page language is set, or what language is specified for a text fragment (but in the latter case it will be a no issue as long as a language is specified correctly).

I'm thinking that it might be good to let users know about differences in behavior listed above, but I'm not sure if we can do it effectively with current UI notification elements at our disposal. We also should, I think, avoid large changes before the release. So maybe we should postpone it for later. Or maybe I'm worrying about it too much. What do you think?


packages/data-models/package.json
```diff
@@ -75,7 +75,8 @@
       "allowImportExportEverywhere": true
     },
     "rules": {
-      "no-prototype-builtins": "off"
+      "no-prototype-builtins": "off",
+      "accessor-pairs": "warn"
```
balmas, line 79: @kirlat the accessor-pairs rule is apparently needed for inflection-tables too with the latest version of the linter. Maybe you can make the change for all of the packages?

kirlat: Sure, will update them

PR #394, opened 2020-05-27
fix #392

### LINKED ISSUE #392: should we add - (hyphen) to list of punctuation?
Right now if a word has a hyphen it gets included in the lookup. E.g.

veni-

I think maybe we should consider hyphen punctuation and strip it out.

balmas:
add hyphen to punctuation lists


balmas:
> Looks good to me!
> I've noticed that a lot of replacement languages are common among languages. I'm thinking that maybe we, in the future, may have a list of "language agnostic" characters in a base class and then merge it with "language specific" characters defined in a model of a specific language? What do you think?

yes that's a very good idea. I'll make a new issue for that.

Issue #385, opened 2020-05-26
Autocomplete broken - Logeion API change


balmas:
it looks like the Logeion API is requiring an API key now. Investigating...

balmas:
ok, we have a key now and the url and the syntax has also changed.

Baseurl for and apikey for logeion requests can now be retrieved from the config service. Settings are described at alpheios-project/alpheios-config-api#3

we can also now add the language to the request so we don't have to filter the wrong language entries out ourselves anymore.

The new syntax is

`<loegion baseurl>?key=<key>q=<query>&lang=<lang>`

balmas:
@irina060981 you can retrieve the logeion and baseurl settings from the UIController.appConfig object (available in state in `api.app.config`). If that object is not defined then the autocomplete feature should be unavailable.

We can eliminate the call to filter out the other language now too, by using the lang parameter.
Let me know if you have any questions. Thanks!

irina060981:
>We can eliminate the call to filter out the other language now too, by using the lang parameter.

@balmas , from my testing we couldn't stop eliminating by language in our code, because logeion API doesn't do it correctly

for example the request:
https://api-v2.logeion.org/logeion2/search?key=<key>&q=male&lang=lat

returns greek words
![image](https://user-images.githubusercontent.com/12377640/83121139-44dde780-a115-11ea-9483-a01a69a86d77.png)

balmas:
![eba4a479-81b2-441d-890d-dffd14984752](https://user-images.githubusercontent.com/644870/83139171-edf6f480-a0b9-11ea-8fe0-b14c1c827219.png)

Yes I see what you mean. Apparently the parameter needs to be set to either 'latin' or 'greek' but that doesn't work either...

irina060981:
merged and closed

balmas:
Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

monzug:
autocomplete is in. tested in Chrome on PC and Mac.
I checked the elimination of words from other languages, but this hasn't been implemented correct?
e.g. car -> carle, cart, Carmo are still in autocomplete

balmas:
>I checked the elimination of words from other languages, but this hasn't been implemented correct?
e.g. car -> carle, cart, Carmo are still in autocomplete

[...]

PR #395, opened 2020-05-28
Add API key to logeion call

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/385

I didn't remove filterAndLimitWords and continued to use because logeion API doesn't filter by lang properly

for example the request:
https://api-v2.logeion.org/logeion2/search?key=<key>&q=male&lang=lat

returns greek words
image


irina060981:
ok, I have  updated the fetch url
will merge it now

Issue #398, opened 2020-05-29
pop-up shrinks and enlarges when downloading to flashcard


monzug:
When downloading a wordlist to flashcard while lexical data message is loading, the pop-up changes in size few times, it shrinks then it enlarges then it shrinks again few times (Chrome on Mac)

balmas:
I can live with the growing/shrinking but the fact that is happening tells me that some state variable is probably being affected by the lookups that is affecting the popup display.  

@irina060981 can you investigate?

irina060981:
yes, I could see this - each word from wordlit update uploads in popup (and it is strange)
will check

irina060981:
merged and closed

balmas:
Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

monzug:
This is not longer reproducible in build  Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

Issue #201, opened 2020-03-17
export wordlists to a format supported by quizlet


balmas:
format details tbd

requested by Matteo Romanello

irina060981:
As I could see - it could be done easily
Import could be done this way:

![image](https://user-images.githubusercontent.com/12377640/78740030-f758c000-7998-11ea-95ba-c217a50562da.png)

irina060981:
![image](https://user-images.githubusercontent.com/12377640/78745214-d1d2b300-79a6-11ea-8b70-83813801d0a5.png)

In quizlet it would be imported without additional tunning

![image](https://user-images.githubusercontent.com/12377640/78745249-e7e07380-79a6-11ea-9c98-53bc342389dd.png)

irina060981:
Done with minimal requirements from #200 
```
word	definition
beatum (beatus, beata, beatior, beatissimus)	happy, fortunate, bringing happiness; rich, wealthy, copious, sumptuous;
beatum (beatus, beata, beatum)	blessed, blissful; "Saint" (in early Church, less formal);
beatum (beo, beare, beavi, beatus)	bless, make happy, gladden, delight; enrich (with);
beatum (beatus, beati)	happy/fortunate men/persons (pl.); "the_rich"; The Blessed, Saints;
beatum (beatum, beati)	happiness, blessedness; good fortune; good circumstances;
mare (mare, maris)	sea; sea water;
mare (mas, maris)	male (human/animal/plant); man;
mare (mas, (gen.), maris)	male; masculine, of the male sex; manly, virile, brave, noble; G:masculine;
mare (mare)	male; masculine, of the male sex; manly, virile, brave, noble; G:masculine;
```

irina060981:
I suggest to close it and continue working on new requirements inside #200 
@balmas , what do you think ?

irina060981:
Changes are merged

balmas:
> I suggest to close it and continue working on new requirements inside #200
> @balmas , what do you think ?

agree

balmas:
@monzug FYI, the basics for this are in Alpheios Components 3.3.0-qa.20200410462

It's not a finished feature though so wait to test until #200 is done.

balmas:
ok, ready to test in Alpheios Components 3.3.0-qa.20200501441

you need to update the wordlist endpoint in your env-webext.js to test:

    'wordlist' : 'https://userapis-dev.alpheios.net/v1/words',

monzug:
[...]

PR #403, opened 2020-06-01
Update wordlist update workflow with new events

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/398

- fix bug with creating unuseful actions for wordlist update

For the issue https://github.com/alpheios-project/alpheios-core/issues/201

- fix bug with updating currentSession, frequency and updateDT for wordlist update


packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -1665,9 +1665,6 @@ If no URLS are provided, will reset grammar data.
   onHomonymReady (homonym, source) {
     // if we have a request for updating wordlist for download then we need this event,
```
balmas, line 1666: we can remove the comment too, right?

packages/wordlist/src/controllers/wordlist-controller.js
```diff
@@ -155,6 +159,21 @@ export default class WordlistController {
     WordlistController.evt.WORDLIST_UPDATED.pub(this.wordLists)
   }
 
+  /**
+   * Responds to a WORDLIST_UPDATE_HOMONYM_READY event by updating a wordlist item for a retrieved Homonym
+   * @param {Homonym} data
+   * Emits WORDITEM_UPDATED and WORDLIST_UPDATED events
+   */
+  onHomonymReadyForWordlistUpdate (data) {
+    console.info('onHomonymReadyForWordlistUpdate', data)
```
balmas, line 168: remove debugging print

packages/components/src/lib/controllers/ui-controller.js
irina060981: yes - removed

packages/wordlist/src/controllers/wordlist-controller.js
irina060981: removed

Issue #406, opened 2020-06-01
flashcard download - exclude header, fix filename


balmas:
I think we should exclude the header row from the flashcard download and also use the file extension .tsv instead of .csv to make it clear that it is tab-separated.

irina060981:
fixed, merged and closed

balmas:
Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

monzug:
tested in Chrome and FF on PC and Chrome on Mac. confirm that file ext is tsv now. Header is gone but on Mac it looks like the first  of the wordlist is replacing the header. see screenshot

balmas:
@monzug I think that's just how the editor is interpreting it, but that's okay. The Quizlet import assumes no header row.  

![quizletimport](https://user-images.githubusercontent.com/644870/83642451-da95ce80-a57c-11ea-8914-3d177c49c24d.png)

monzug:
ok. I'll close it

Issue #404, opened 2020-06-01
Character display issue with syriac words in wordlist


monzug:
see issue #349.
I was able again to reproduce the character display issue in wordlist. Go to chrome and activate alpheios in Loeb classics.  double clcik on a couple of words then go to the syriac site sedra and change page language to syriac. double click on few words there. the words are saved in wordlist with character display issue see screenshot
![Screen Shot 2020-06-01 at 3 09 02 PM](https://user-images.githubusercontent.com/41396793/83412536-870b6f80-a41a-11ea-859c-16ddfcf927ec.png)

monzug:
also in morphology pop-up
![Screen Shot 2020-06-01 at 3 16 03 PM](https://user-images.githubusercontent.com/41396793/83412762-f97c4f80-a41a-11ea-9844-e76072ceb93b.png)

and wordlist limit to this session
![Screen Shot 2020-06-01 at 3 16 38 PM](https://user-images.githubusercontent.com/41396793/83412796-06993e80-a41b-11ea-9ef6-0a7f4b4ac908.png)

irina060981:
I was not able to reproduce - but I believe that this problem could be because of using not specific font
I have added lang attribute to wordlist to use a correct font

as for popup font problem - I believe we have some `ara` text there (I saw similiar for other words)

Finally I think that the problem coud be in the case that specific fonts was not loaded correctly.
We upload fonts from 
fonts.gstatic.com

May be there are some problems sometimes with uploading from there?
@balmas and @monzug , what do you think?

monzug:
could be. I was working from my other location where the connection is not
too good.

balmas:
>May be there are some problems sometimes with uploading from there?

Yes that seems possible.

monzug:
let me try again

monzug:
this is chrome on Mac:

![Screen Shot 2020-06-02 at 2 09 35 PM](https://user-images.githubusercontent.com/41396793/83518537-d9ac6080-a4da-11ea-8b85-02aa53a62bb5.png)

![Screen Shot 2020-06-02 at 2 10 03 PM](https://user-images.githubusercontent.com/41396793/83518530-d6b17000-a4da-11ea-9161-5c9c172805a0.png)

balmas:
Chrome on Mac definitely has character encoding problems. Does the same problem occur on the page on which the words were looked up?

monzug:
Chrome on PC - this is book Peshitta, acts of the apostles

[...]

PR #407, opened 2020-06-02
remove header and change download file extension to .tsv

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/406

- removed headers
- change extension file to `.tsv`

For the issue https://github.com/alpheios-project/alpheios-core/issues/408

- fixed punctuation list for syriac model

For the issue https://github.com/alpheios-project/alpheios-core/issues/404

- add lang attribute to wrditem component to use a correct font


packages/data-models/src/syriac_language_model.js
```diff
@@ -90,7 +90,7 @@ for the current node
    * @returns {string} a string containing valid puncutation symbols
    */
   static getPunctuation () {
-    return "\u0700\u0701\u0702\u0703\u0704\u0705\u0706\u0707\u0708\u0709\u070A\u070B\u070C\u070D\u070F-.,;:!?'\"(){}\\[\\]<>/\\\u00A0\u2010\u2011\u2012\u2013\u2014\u2015\u2018\u2019\u201C\u201D\n\r\u200C\u200D"
+    return "\u0700\u0701\u0702\u0703\u0704\u0705\u0706\u0707\u0708\u0709\u070A\u070B\u070C\u070D\u070F\\-\\.,;:!?'\"(){}\\[\\]<>/\\\u00A0\u2010\u2011\u2012\u2013\u2014\u2015\u2018\u2019\u201C\u201D\n\r\u200C\u200D"
```
balmas, line 93: Can you apply the same fix this to all of the language model classes? Thank you.

packages/components/src/vue/components/word-list/download-confirmation.vue
```diff
@@ -175,8 +175,8 @@
           dataForDownload = this.prepareDownloadListFull()
         }
 
-        const result = Download.collectionToCSV(dataForDownload.delimiter, dataForDownload.exportFields)(dataForDownload.wordlistData)
-        Download.downloadBlob(result, `wordlist-${this.languageCode}.csv`)
+        const result = Download.collectionToCSV(dataForDownload.delimiter, dataForDownload.exportFields, false)(dataForDownload.wordlistData)
+        Download.downloadBlob(result, `wordlist-${this.languageCode}.tsv`)
```
balmas, line 179: Hmm. This changes it for both downloads, right? The tsv extension only really applies for the flashcard download though.  It should stay csv for the other.

packages/data-models/src/syriac_language_model.js
irina060981: ok - done

packages/components/src/vue/components/word-list/download-confirmation.vue
irina060981: you are quite right - I missed it! thank you for the catch
fixed

irina060981:
@balmas, as fixes are not big - I will merge it after I fixed them

Issue #402, opened 2020-05-29
cleanup console messages


balmas:
A number of console calls have snuck back in and are cluttering up the console. We should use our Logger library methods instead so that they are only output in verbose mode.

```
./lib/selection/media/html-selector.js:        console.warn(`Alpheios word selection error - no word separator function found for "${textSelector.model.baseUnit.toString()}" base unit`)
./lib/selection/media/html-selector.js:      console.warn('Browser does not support the Alpheios word selection code. Support for getSelection() or createTextRange() is required.')
./lib/options/options.js:                  console.warn(`Unable to parse Alpheios option value for  ${parsedKey.name} from ${values[parsedKey.name]}`, e)
./lib/options/options.js:              console.warn(`Unable to parse Alpheios option value for  ${parsedKey.name} from ${values[parsedKey.name]}`, e)
./lib/options/options.js:      console.warn(`Failed to parse stored Alpheios options key ${key}`)
./lib/utility/html-page.js:      console.warn(`Alpheios cannot be activated on a document without a body element (${window.document.URL})`)
./lib/utility/html-page.js:      console.warn(`Alpheios cannot be activated on a window narrower than ${HTMLPage.targetRequirements.minWidth} (${window.document.URL})`)
./lib/utility/html-page.js:      console.warn(`Alpheios cannot be activated on a window shorter than ${HTMLPage.targetRequirements.minHeight} (${window.document.URL})`)
./lib/utility/html-page.js:      console.warn(`Alpheios cannot be activated on a page with fewer than ${HTMLPage.targetRequirements.minCharCount} (${window.document.URL})`)
./lib/utility/platform.js:      console.warn('Alpheios cannot determine the orientation of this device, assuming "portrait"')
./lib/utility/platform.js:      console.warn('Alpheios cannot determine what platform this is - either document or documentElement are not defined')
./vue/components/lookup.vue:        console.warn(`Lookup request cannot be completed: ${err.message}`)
./vue/vuex-modules/data/lexis.js:      console.warn('CEDICT functionality will be disabled because LexisCS configuration is not available')
./vue/vuex-modules/data/lexis.js:          console.warn(err.message)
./vue/vuex-modules/data/lexis.js:          console.warn(err.message)
./vue/vuex-modules/data/lexis.js:      console.warn('Lookup request cannot be completed: LexisCS configuration is unavailable')
./vue/vuex-modules/data/lexis.js:          console.warn(err.message)
[...]

PR #409, opened 2020-06-02
Updated logging

kirlat:
Changed console logs to use Alpheios logger instead, for #402.


irina060981:
I have a question about the code:
 - sometimes in the code you use `Logger.getInstance()`
 - sometimes you initialize and creates one instance and use `this.logger`
 - sometimes you use `the.$options.logger`

may be it is better to use it one way?
as I could see in the code - we should always have a single instance  that is defined inside `data-models`  library `Logger` class
so this way it is worth to use  `Logger.getInstance()` without any aditional  saving to UIController instance and to `options` module instance

packages/components/tests/lib/options/options.test.js
```diff
@@ -1,19 +1,17 @@
...
 import Options from '@/lib/options/options'
 // import OptionItem from '@/lib/options/options-item'
 import StorageAdapter from '@/lib/options/storage-adapter'
 import LocalStorageArea from '@/lib/options/local-storage-area'
 
 describe('options.test.js', () => {
-  console.error = function () {}
-  console.log = function () {}
-  console.warn = function () {}
+  const logger = Logger.getInstance({ verbose: true })
```
irina060981, line 10: As I could see finally logger use the same console commands, 
and this stub construction should continue to work (if verbose mode is defined right)

This way I believe that it is enough in `beforeAll` block define logger inside each test with verboseMode
and tests should work well
And there is no big value in redefining the whole tracking workflow

We have this tracking construction in the most of all tests in all packages, I think
So I believe the change with defining logger should be done in each of them

kirlat:
> I have a question about the code:
>     * sometimes in the code you use `Logger.getInstance()`
>     * sometimes you initialize and creates one instance and use `this.logger`
>     * sometimes you use `the.$options.logger`
> may be it is better to use it one way?
> as I could see in the code - we should always have a single instance that is defined inside `data-models` library `Logger` class
> so this way it is worth to use `Logger.getInstance()` without any aditional saving to UIController instance and to `options` module instance

I agree that `Logger.getInstance()` is the best way to use logger across the code. I, however, was trying to touch an existing code as little as possible before there release. That's why I left other usages of it that existing before my changes. 

@balmas: I can do this change now, or can create an issue and wait until after the release. What do you think?

packages/components/tests/lib/options/options.test.js
kirlat: I was checking `Logger.log()` calls instead of `console.log()` to make sure it goes through our logger and does not bypass it, since the requirement is to always use our logger, not the `console`. But if we would check for the `console` calls, we would skip that check and calls to `console` functions might sneak it. Does it sound reasonable?

irina060981: For now with your code we have the following problem:
if we won't define Logger with verbose mode: true, we could fail all tests that have something like the following

`expect(console.error).toHaveBeenCalledWith(expect.stringMatching(/storing/))`
`expect(console.warn).not.toHaveBeenCalledWith(expect.stringMatching(/IndexedDB/))`

>I was checking Logger.log() calls instead of console.log() to make sure it goes through our logger and does not bypass it, since the requirement is to always use our logger

I believe that it is worth to create a separate test that would should show
that with different verbose modes we could catch console.warn and console.error in some one module
And it is not very consistent to change the overall workflow of one test file for this purpose
but of course it is from my point of view :)

balmas:
> > I have a question about the code:
> > * sometimes in the code you use `Logger.getInstance()`
> > * sometimes you initialize and creates one instance and use `this.logger`
> > * sometimes you use `the.$options.logger`
> > may be it is better to use it one way?
> > as I could see in the code - we should always have a single instance that is defined inside `data-models` library `Logger` class
> > so this way it is worth to use `Logger.getInstance()` without any aditional saving to UIController instance and to `options` module instance
> 
> I agree that `Logger.getInstance()` is the best way to use logger across the code. I, however, was trying to touch an existing code as little as possible before there release. That's why I left other usages of it that existing before my changes.
> 
> @balmas: I can do this change now, or can create an issue and wait until after the release. What do you think?

Create an issue and wait until after the release. Thank you.

packages/components/tests/lib/options/options.test.js
kirlat: > if we won't define Logger with verbose mode: true, we could fail all tests that have something like the following

From my point of view (and I can be wrong) if any test is checking calls to `console.error()` within `Logger.error()` they should switch to checking for `Logger.error()` calls instead. The knowledge that `Logger.error()` is using `console.error()` underneath are internal implementation details. Tests cases should not rely on the knowledge of internal implementation details of the class not being tested (and, on my opinion, internal details of a class being tested should not be accessed directly too, but there are different approaches and there are exceptions).

> And it is not very consistent to change the overall workflow of one test file for this purpose

I agree that it breaks the consistency across test cases. So we might switch to testing for `console` calls everywhere for it to be consistent. It might be worth it.

There are many different approaches to testing, and each has its own benefits. It's often hard to make an objective decision about which one to use. I do not have a very strong opinion about one or the other. I think Irina's point about consistency in test suits is very good, and it might make sense to sacrifice purity for that. @balmas, what do you think? What should be our approach in this case?

Issue #417, opened 2020-06-03
fix the size of the header when repositioning


balmas:
Reported by @abrasax 

If the popup is taller than the screen, when we are calculating the correct position for it there is a phase when the popup header gets smooshed before it returns to its normal size.

This can be seen in the video at 
https://www.awesomescreenshot.com/video/180814?key=3acad7792b77c2f89a9781c26e45b219

@kirlat can we fix this in the CSS somehow?

kirlat:
A fix is in the master now.

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200605325
Alpheios Components 3.3.0-qa.20200605311

monzug:
I knew Harry wouldn't like it! fix in latest build  Alpheios Reading Tools 3.3.0 build qa.20200605325
Alpheios Components 3.3.0-qa.20200605311 
tested in Chrome, FF and Safari

PR #419, opened 2020-06-04
A popup's header fix

kirlat:
Fixes an issue with the size of a popup's header during resizing (#417)


packages/components/src/vue/components/popup.vue
```diff
@@ -275,6 +275,8 @@ export default {
           return `${this.resizedHeight}px`
         }
         this.$options.logger.log(`${time}: height getter, return value is ${this.heightValue}`)
+        const value = this.heightValue === 'auto' ? 'auto' : `${this.heightValue}px`
+        console.info(`height is ${value}`)
```
balmas, line 279: remove the debugging print.

kirlat: Oh, yes, thanks for catching this!

Issue #422, opened 2020-06-04
cannot select same word twice in a row with Long Tap on Mobile


balmas:
This appears to be a new problem with this release -- I can't reproduce it at texts.alpheios.net, but can reliably at texts-test.alpheios.net

Steps to reproduce:

https://texts-test.alpheios.net/text/urn:cts:latinLit:phi0959.phi001.perseus-lat2/passage/1.ep.1-1.ep.4

long tap on "modo"
see the popup appear
close the popup
long tap on "modo" again
no popup
long tap on another word
popup appears
long tap again on "modo"
popup appears

Same sequence at https://texts.alpheios.net/text/urn:cts:latinLit:phi0959.phi001.perseus-lat2/passage/1.ep.1-1.ep.4 does not demonstrate the problem.

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200609446
Alpheios Components 3.3.0-qa.20200609405

monzug:
I can't repro the issue described above on my iPhone xr.

PR #423, opened 2020-06-05
A fix for word selection on mobile

kirlat:
Fixed an issue with not being able to select the same word twice on mobile. See #422 for more details.


kirlat:
Thanks for catching this issue! This was a very intricate one and was appearing under certain conditions on mobile only where the popup did not exist and, as a result, `isPopupVisible()` was throwing an exception. Now, with the additional checks added, we'll have a code that will be able to handle such cases (and probably some other too) much more reliably.

Issue #426, opened 2020-06-05
treebank icon is available for lookup word that is not in the diagram


monzug:
while testing #166, I have noticed that if I enter a latin word in lookup (that is NOT in the diagram) such as caro, the pop-up will offer the diagram icon. when I click on it, it will generate the last seen diagram
to repro: in https://texts-test.alpheios.net/text/urn:cts:latinLit:phi0959.phi006.alpheios-text-lat1/passage/1.1-1.30
double click on mundi and look at the diagram. the close pop-up, go to lookup and enter caro. the pop-up will have the diagram icon and will generate the diagram with mundi.
this is not reproducible in release environment

![caro-with-diagram](https://user-images.githubusercontent.com/41396793/83890967-8163a180-a71a-11ea-8386-d0bd011818f0.JPG)

the icon is not longer generated after I double click on a word with no diagram

balmas:
@kirlat is it possible we broke this with latest changes for #258 (for https://github.com/alpheios-project/alpheios-core/issues/258#issuecomment-634119573) ?

kirlat:
Will look into that

kirlat:
> @kirlat is it possible we broke this with latest changes for #258 (for [#258 (comment)](https://github.com/alpheios-project/alpheios-core/issues/258#issuecomment-634119573))

Yes, that was it and then another. Submitted a fix.

kirlat:
Code that solves this is in master now

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200609446
Alpheios Components 3.3.0-qa.20200609405

monzug:
Fixed. Thanks

PR #427, opened 2020-06-05
A treebank icon fix

kirlat:
Fixes an issue with a treebank icon shown when treebank data is not available, see #426


packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -299,6 +301,8 @@ export default class Lexis extends Module {
         store.commit('lexis/setTreebankInfo', { hasTreebankData: false })
       }
     } else if (this._treebankAvailable) {
+      // TODO: Do we need to disable an icon?
```
balmas, line 304: is there still a TODO here?

kirlat: Oh, yes, will remove. Sorry

Issue #432, opened 2020-06-08
Super minor: loeb classics button has wrong background


monzug:
FF on MAC
see screenshot 
![Screen Shot 2020-06-08 at 3 57 06 PM](https://user-images.githubusercontent.com/41396793/84038986-e0d2e300-a9a0-11ea-9ad5-1c7995bdb8a2.png)

kirlat:
A fix has been merged to master.

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200609446
Alpheios Components 3.3.0-qa.20200609405

PR #433, opened 2020-06-08
Loeb classics fixes

kirlat:
Fixes an issue with a button text color on Loeb classics (see #432).

Issues like this makes me want to have an ability to isolate styles of components from those of a page as in Shadow DOM.


balmas:
>Issues like this makes me want to have an ability to isolate styles of components from those of a page as in Shadow DOM.

Yes me too

Issue #382, opened 2020-05-26
date stamps gets updated with today's date in wordlist


monzug:
I did notice first in FF in loebclassics as test user. when logged in for the first time, most of the words in my saved wordlist had today's date. then I logged out and logged with sign in with apple in Latinlibrary. as soon as I logged in, words in my wordlist were saved with today's date. the only exception are the words marked as important, they have a different date (I am assuming that that's the date when they were marked as important, correct? or should it be when they were added?)
see attachment
![Screen Shot 2020-05-26 at 2 42 23 PM](https://user-images.githubusercontent.com/41396793/82902387-abb9a000-9f5f-11ea-9549-7973bc779090.png)

monzug:
see more examples of this issue
![today'sdate](https://user-images.githubusercontent.com/41396793/83755370-7be35a00-a63b-11ea-80c5-a1bbcbb2a68d.JPG)

balmas:
Assigning this to @irina060981 to investigate. 

What I think should happen is the following:

1) Existing words in the wordlist (i.e. those that were added with the prior release of the code) should only get date stamps in the following circumstances
a) they are looked up again (from either the page or the wordlist)
b) their important status is changed

2) Words added to the wordlist with this release of the code should get the date stamp of the day they were added. Their date stamps should be updated in the same circumstances as above, i.e. 
a) they are looked up again (from either the page or the wordlist)
b) their important status is changed

balmas:
@monzug actually can you confirm how what you are seeing differs from that?

monzug:
yes, when I limit to most recent  and I have added only 4 words to wordlist
in all day for the user (the limit is to 25 words),  I get 25 words with
date stamp updated with the TODAY date if wordlist is long. maybe this will
be automatically fixed by this fix.
also, how are these extra words picked from existing wordlist? it looks
like it's alphabetical order but it should be most recent?

balmas:
@irina060981 can you investigate this one?

irina060981:
If to be honest I couldn't reproduce it with non-testuser:
I logged with my account:

![image](https://user-images.githubusercontent.com/12377640/84098776-77f47500-aa4b-11ea-9f60-6a52cc6ad9a5.png)

I have a got a long word list created 06/01
I downloaded with flashcard check without filtering and it did lexical request

![image](https://user-images.githubusercontent.com/12377640/84098849-a2dec900-aa4b-11ea-8a37-ff61e20e8ba9.png)

[...]

Issue #309, opened 2020-05-07
new "Advanced" options tab


balmas:
@abrasax and I discussed this and agreed that we would like to have a new "Advanced" options tab, which contains the mousemove and google docs settings. We should also put the Log Level setting there as well.

@irina060981 I understand your concern about it being a bucket of "other" settings, but we think that's better than having them in the UI tab.

irina060981:
Merged and closed

balmas:
Alpheios Components 3.3.0-qa.20200513442

monzug:
question: we still have Mouse Move in Advanced tab but mouse over in Word Selection: Use Mouseover to Select Words (recommended for Chinese)

balmas:
I will look at all the mouseover text and regularize it.

irina060981:
I am sorry, but I couldn't explain why it is again Mouseover here
from my memory - I changed it twice times, but it continues to be Mouseover
and again I couldn't find this change in my PRs :)

it is some tricky games with my mind!

balmas:
Alpheios Reading Tools 3.1.0 build qa.20200602407
Alpheios Components 3.3.0-qa.20200602383

Normalization of mousemove language. Should now say "Mouse movement over text"

monzug:
YES! mouse movement is now used in Features and in Advanced tab.
as per mobile, is there anything that can be done to improve the view of the advanced tab in iPhone 8 and iPhone xr?  in iPhone 11 they are all aligned
![IMG_4547 (1)](https://user-images.githubusercontent.com/41396793/83763013-34ae9680-a646-11ea-9d69-514faf239581.PNG)

balmas:
@irina060981 can you look at the formatting on mobile?

irina060981:
I suggest to reduce the font-size this way

![image](https://user-images.githubusercontent.com/12377640/84140022-481d8f80-aa94-11ea-89ba-831fcbafa1f1.png)

irina060981:
merged and closed

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200609446
Alpheios Components 3.3.0-qa.20200609405

monzug:
Looks good now, except when going from Advanced to Resources, the Resources tab is shift to the left. UI, Features and Advanced are displayed correctly.

balmas:
I'm guessing this should be an easy css change...

monzug:
[...]

PR #436, opened 2020-06-09
Fixes for wordlist workflow

irina060981:
For the https://github.com/alpheios-project/alpheios-core/issues/382

- fix for the correct upload for the data from remote 
  it also fixed the disapperaing of lemmaList 

added a fix for https://github.com/alpheios-project/alpheios-core/issues/309

I suggest to reduce the font-size this way

image


packages/data-models/src/homonym.js
```diff
@@ -65,13 +65,18 @@ class Homonym {
...
       for (const lexeme of jsonObject.lexemes) {
         lexemes.push(Lexeme.readObject(lexeme))
       }
+    } else {
+      const languageID = LMF.getLanguageIdFromCode(jsonObject.languageCode)
+      lexemes = [new Lexeme(new Lemma(jsonObject.targetWord, languageID), [])]
     }
-    const homonym = new Homonym(lexemes, jsonObject.form)
+    const homonym = new Homonym(lexemes, jsonObject.form || jsonObject.targetWord)
+    homonym.lemmasList = jsonObject.lemmasList
```
irina060981, line 79: we need this because we retrieve homonym data from remote as 
{ targetWord, lemmaList } without lexemes property

packages/wordlist/src/controllers/user-data-manager.js
```diff
@@ -280,8 +280,9 @@ export default class UserDataManager {
       } else {
         remoteItems = await remoteAdapter.query(data.params)
         if (params.type === 'full') {
-          for (let remoteItem of remoteItems) {
-            await localAdapter.checkAndUpdate(remoteItem, data.params.segment, [remoteItem])
+          for(let remoteItem of remoteItems) {
+            let wodrItem = localAdapter.dbDriver.createFromRemoteData(remoteItem)
+            await localAdapter.checkAndUpdate(wodrItem, data.params.segment, [remoteItem])
```
irina060981, line 285: we need to create wordItem first and then pass it to localAdapter for update

packages/data-models/src/homonym.js
balmas: can you add a comment that says this to the code?

packages/wordlist/src/controllers/user-data-manager.js
```diff
@@ -280,8 +280,9 @@ export default class UserDataManager {
       } else {
         remoteItems = await remoteAdapter.query(data.params)
         if (params.type === 'full') {
-          for (let remoteItem of remoteItems) {
-            await localAdapter.checkAndUpdate(remoteItem, data.params.segment, [remoteItem])
+          for(let remoteItem of remoteItems) {
+            let wodrItem = localAdapter.dbDriver.createFromRemoteData(remoteItem)
```
balmas, line 284: minor point, but maybe fix the typo in the variable? "wordItem" instead of "wodrItem" ?

PR #439, opened 2020-06-09
fix for #382

balmas:
only update wordlist updatedDT and frequency when an actual lookup occurs
not when doing a lookup for wordlist download

Requesting review from both @kirlat and @irina060981  because I would like to get the fix in for a build today but also want @irina060981 to make sure she doesn't see a problem with it.

This fixes 2 problems:

1) wordlist items were automatically getting updatedDT and frequency fields updated when the local db was synced with the remote, because the fix for that in f071aba445b118a383b36a9dfee9652dd4857c7d only applied to items pulled from remote and missed the sync with the local.

2) wordlist items that were downloaded for flashcards and for which a lookup was needed to pull the definition were getting the updatedDT and frequency updated in the save to the remoteDB. These items should be updated for short definition only.


irina060981:
@balmas , I have added and not removed later this asignment only because 
we were going to make these fields required in the remote source.
Because we shouldn't have situations when these field could be empty on saving to source.

with this fix we won't be able to change the requirements to data sources, I believe

balmas:
> @balmas , I have added and not removed later this asignment only because
> we were going to make these fields required in the remote source.
> Because we shouldn't have situations when these field could be empty on saving to source.
> 
> with this fix we won't be able to change the requirements to data sources, I believe

True. I think that will have to wait for the next release, when perhaps we can notify people that we are going to update their wordlists.

Issue #445, opened 2020-06-10
wordlist sorting problems in Safari extension


balmas:
When you have a list with a mixture of words with and without updatedDate, the sort by most recent in Safari does not work as expected:

See also 
https://github.com/alpheios-project/alpheios-core/issues/382#issuecomment-642049027

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238

monzug:
sorting looks good in latest safari build Safari App Extension 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238

Issue #444, opened 2020-06-10
wordlist: I can add a failed word to wordlist but it does not show up in limit to most recent


monzug:
in http://www.thelatinlibrary.com/petronius1.html, I double clicked on Thucydidis and got a lexical query produced no results. the word is added to the wordlist. see current session screenshot

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238

monzug:
can't reproduce in latest build.
![compensarium](https://user-images.githubusercontent.com/41396793/84401793-adc36600-abd1-11ea-8f32-bc25d54e4852.JPG)

Issue #443, opened 2020-06-10
currentSession, frequency and updatedDate not being updated when lookup is from wordlist


balmas:
Double click on a word on the wordlist, view popup, and note that currentSession, frequency and updatedDate have not channged.

_Originally posted by @monzug in https://github.com/alpheios-project/alpheios-core/issues/430#issuecomment-642025990_

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238

monzug:
see dextris and Florus both been updated from wordlist, github #430 
![destris-florus](https://user-images.githubusercontent.com/41396793/84411825-6b535680-abdc-11ea-9867-686ff4037a0c.JPG)

Issue #430, opened 2020-06-08
wordlist: current session icon added only for failed lookup


monzug:
While testing the limit to download with flashcard, I have noticed the following:
only for the failed lookup words, the current session icon is added to wordlist
see screenshot
![Screen Shot 2020-06-08 at 2 46 51 PM](https://user-images.githubusercontent.com/41396793/84031998-f0e5c500-a996-11ea-9bbd-c8335649ec78.png)

to repro: 
limit to 8 the number of words for flashcard export in Options
go to latin wordlist (with more than 8 words in the list) for any existing user, then
download to flashcard and check the downloaded file.
as there are only 6 words in the downloaded file, I checked which words had a failed lookup by double clicking from wordlist on each of the 8 words. the two words with failed lookup (in my scenario these are Florus and Options) added the current session icon and the most often # increased to 2 after double clicking on them. the other 6 words did not get the current session icon or increased the most often number. Why such a different behavior?

balmas:
sounds like an oversight in the event handling. Assigning to @irina060981

irina060981:
I couldn't reproduce it either
when I double click  on the word that is already exists inside wordlist - it gets currentSession icon and frequency is updated (and it doesn't depend on if it has lexical query result)

![image](https://user-images.githubusercontent.com/12377640/84138808-418e1880-aa92-11ea-9b8b-b562fa6cd2fb.png)

if I would clear local IndexDB data, reload the page (in this case wordItems doesn't have definitions inside homonym), and click flashcard download, it does update from lexical query but it doesn't change currentSession for any word

irina060981:
@balmas , could you reproduce it?
by the way, I checked with last changes from https://github.com/alpheios-project/alpheios-core/pull/436
but I don't think that it has dependings

balmas:
I am hoping this is fixed with the latest changes in #441 for #166

balmas:
Alpheios Reading Tools 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238

monzug:
not longer reproducible in  Alpheios Reading Tools 3.3.0 build qa.20200611246
Alpheios Components 3.3.0-qa.20200611238 
dextris gets the current session + most oftern # updated
![destris-florus](https://user-images.githubusercontent.com/41396793/84410662-0dbf0a00-abdc-11ea-8095-b9be1276b482.JPG)

PR #447, opened 2020-06-10
i443 and i445

### LINKED ISSUE #442: words from current session missing in limit by most recent
Chrome on PC. login as google user monzug again
1) add some new words to wordlist (Alpibus, venturum, copiis), go to wordlist and limit by this session. see screenshot (all kosher here)

session-limit

2) then limit by most recent. note that the 3 words are missing from the most recent list in wordlist. see screenshot

most-recent-limit

_Originally posted by @monzug in https://github.com/alpheios-project/alpheios-core/issues/430#issuecomment-642025990_

balmas:
this PR includes my fix for #443 and @kirlat's fix for #445 
I believe it also fixes #442 , #444 and #430


.travis.yml
```diff
@@ -26,8 +26,8 @@ deploy:
   - provider: script
     script: "npx lerna publish from-package --yes --dist-tag rc"
     on:
-      branch: production
-      tags: true
+      all_branches: production
+      tags: on
```
kirlat, line 30: I'm not sure about these changes. I don't know what our goal is, but the format seems to contradict Travis docs: https://docs.travis-ci.com/user/deployment#conditional-releases-with-on.
According to the docs, the only possible value for `all_branches` is `true` and `tags` can be either `true`, `false`, or a tag name.

balmas: oh, sorry, this was not meant to be committed, I had been testing different things with the production builds. I will revert this before merging.

Issue #458, opened 2020-06-16
Update intl-messageformat


kirlat:
We're currently using an outdated version of `intl-messageformat` package, 2.x.x. Since then it has become a part of a [Format.js](https://formatjs.io/) multi-package repository. We should update our code to use the latest version of the library.

kirlat:
It seems that Intl Message Format 8.x.x does not like attributes of HTML tags within messages. It now checks for something that does not match "plain" XML tags in the message and throws an error if anything is found.

Luckily, only two messages across all the ones that we use are affected:
```
  "TEXT_NOTICE_CHANGE_LANGUAGE": {
    "message": "<span lang='{langCode}'>'{targetWord}'</span> was not found in our {languageName} resources. Wrong language? Change to:",
    "description": "language notification",
    "component": "UI",
    "params": ["targetWord", "languageName", "langCode"]
  },
  "TEXT_NOTICE_NOT_FOUND": {
    "message": "<span lang='{langCode}'>'{targetWord}'</span> was not found in our {languageName} resources.",
    "description": "language notification",
    "component": "UI",
    "params": ["targetWord", "languageName", "langCode"]
  }
```
@balmas, @irina060981: do we really need this `langCode` attribute there? What is it for? I can't see any purpose of it from the top of my head.

Can we replace the above with something like
```
  "TEXT_NOTICE_CHANGE_LANGUAGE": {
    "message": "'{targetWord}' was not found in our {languageName} resources. Wrong language? Change to:",
    "description": "language notification",
    "component": "UI",
    "params": ["targetWord", "languageName"]
  },
  "TEXT_NOTICE_NOT_FOUND": {
    "message": "'{targetWord}' was not found in our {languageName} resources.",
    "description": "language notification",
    "component": "UI",
    "params": ["targetWord", "languageName"]
  }
```
?

I think that if we remove a `lang` attribute `<span>` is not needed any more?

kirlat:
FYI, there seems to be a feature of tag replacement available: https://formatjs.io/docs/intl-messageformat#rich-text-support

I don't know if we need it right now, but maybe there will be some use for it in the future.

kirlat:
Also, FYI, tags in messages has to be XHTML compatible. As a result, we cannot use some HTML5 tags such as `<br>`, it has to be `<br/>` instead. `<br>` would be considered as an unclosed tag.

balmas:
[...]

PR #461, opened 2020-06-17
IntlMessageFormat update

kirlat:
This is an update of `intl-message` format to its latest version and related format and code changes. I've also added a couple of lines to improve error handling in `L10n` module. I've also updated a majority of other dependencies to their latest versions, except for `uuid` which causes some issues in a test environment. We can probably postpone `uuid` update to later. For #458.


packages/data-models/src/language_model.js
```diff
@@ -420,12 +420,18 @@ class LanguageModel {
...
+   * @param {string} preceding preceding word (optional)
+   * @param {string} following following word (optional)
+   * @param {string} encoding encoding name to filter the response to (optional)
+   * @param {boolean} preserveCase if true will preserve the case (default is false)
+   * @param {boolean} includeOriginal if true will include the original word even if it is unchanged (default is false)
+   * @param params.preceding
+   * @param params.following
+   * @param params.encoding
+   * @param params.preserveCase
+   * @param params.includeOriginal
```
balmas, line 434: I think the linter didn't understand our js docs here  because it looks like it repeated them

kirlat: Oh, I've missed that. I think the problem is that we declared parameters as independent variables and linter wanted them to be props of the `params` object. I will correct a JSDoc block.

Issue #464, opened 2020-06-18
Lookup button not working in mobile


balmas:
reported by @monzug 

If you paste or enter a word in the lookup box on mobile and then touch the Lookup button, nothing happens. using the keyboard "Go" or enter succeeds. It seems the handler on the button isn't using the right event.

monzug:
it does work if I enter a word in lookup, only paste a word does not work on iPhone 11 and XR

balmas:
it didn't for me though, at least it didn't if the word wasn't long enough to bring up the autocomplete. Once the autocomplete dropdown shows, it suddenly worked.

Maybe it's the autocomplete functionality that is causing the problem then...

monzug:
it worked for me with oi (no autocomplete)

balmas:
hmm. not for me on a Samsung Galaxy 7.

But the autocomplete might be a red-herring, turning the it off in options doesn't change anything.

However, interestingly, once I successfully submit the lookup with the 'Go' button on my keyboard, if I leave the lookup box open, I can then use the button to submit additional words. Once I close the box and reopen it though, the problem recurs.

balmas:
this is similar to a problem I was seeing intermittently on the desktop too -- the initial click on the Lookup button failed, subsequent uses of it succeeded.

kirlat:
Hmm, I've tested it on my phone on `texts-text.alpheios.net` in Chrome, and I was not able to find any single case when it fails so far. No matter whether I type the word in or select it from a list, it always starts a lookup. Will keep trying.

balmas:
@kirlat it starts the lookup after a press on the Lookup button?

balmas:
what if you copy/paste it in, instead of entering it?

kirlat:
> @kirlat it starts the lookup after a press on the Lookup button?

Yes (the orange one with the "Lookup" word on it to be absolutely sure)

kirlat:
> what if you copy/paste it in, instead of entering it?

If I copy and past it, it does not work for me. But if I type a new word after that, it does.

balmas:
ok, then that is at least similar to what I am seeing.

monzug:
yes, same here

kirlat:
[...]

PR #465, opened 2020-06-18
A lookup paste fix

kirlat:
A fix of an issue with a paste event not being handled properly by the lookup component (see #464).


packages/components/src/vue/components/form-components/input-autocomplete.vue
```diff
@@ -92,6 +93,11 @@ export default {
       this.getAutocompleteWords()
     },
 
+    onPaste (event) {
+      this.valueText = (event.clipboardData || window.clipboardData).getData('text').trim()
```
balmas, line 97: could both event.clipboardData and window.clipboardData ever be null? It might be safer to make sure that they aren't before calling getData otherwise we'll throw an exception here.

kirlat: This is a good suggestion, will add that check.

PR #481, opened 2020-06-26
UI controller refactoring phase 3

kirlat:
In this step, I moved a majority of UI-related functions into a UI controller. It will require an update of webextension and embed-lib. I will submit a PR for those once this PR will be merged.

I understand that it might be hard to review the changes because there will be no comparison shown between an old and a new version. But I think maybe you will be able to check for some tricky conditions and/or checks, or any specific business logic cases that I might have twisted inadvertently. It's good to be sure that there are no unexpected changes caused by this update.

In the next step I'm planning to move a `ui` module of Vuex into the UI controller. I will also review a remaining UI-related functionality in the App controller and will decide what can be moved to the UI controller and what can be simplified or eliminated.

Thanks!


packages/components/src/lib/controllers/app-controller.js
```diff
@@ -968,6 +956,15 @@ if you want to create a different configuration of an app controller.
     // Inject Alpheios CSS rules
     this.activateOnPage()
 
+    this.isActivated = true
+    this.isDeactivated = false
+
+    this.activateModules()
+    // this.state.activate()
```
balmas, line 963: why is this commented out?

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1934,15 +1713,15 @@ If no URLS are provided, will reset grammar data.
...
+  /* uiSetFontSize (uiOptions) {
     const FONT_SIZE_PROP = '--alpheios-base-text-size'
     try {
       document.documentElement.style.setProperty(FONT_SIZE_PROP,
         `${uiOptions.items.fontSize.currentValue}px`)
     } catch (error) {
       this.logger.error(`Cannot change a ${FONT_SIZE_PROP} custom prop:`, error)
     }
-  }
+  } */
```
balmas, line 1724: remove commented method?

packages/components/src/lib/controllers/ui-controller.js
```diff
@@ -57,39 +82,77 @@ export default class UIController {
...
     this.activateModules()
+    this._uiState.activateUI()
+
+    if (this.hasModule('panel')) {
+      if (this._uiState.isPanelStateDefault() || !this._uiState.isPanelStateValid()) {
+        this._uiState.setPanelClosed()
+      }
+      if (this._uiState.isPanelOpen()) {
+        this.openPanel(true)
+      }
```
balmas, line 141: this logic is a little different than before. is there any chance the panel will be left in an undefined state?

packages/components/src/lib/controllers/app-controller.js
kirlat: I just forgot to remove it. It has been moved to the UI controller and is no longer needed: https://github.com/alpheios-project/alpheios-core/blob/dev-ui-controller-refactoring-3/packages/components/src/lib/controllers/ui-controller.js#L131. Will clean it up

packages/components/src/lib/controllers/app-controller.js
kirlat: Yes, just forgot about this one.

packages/components/src/lib/controllers/ui-controller.js
kirlat: I thought it is an equivalent but I might be missing something.
In master, it is:
```
    if (this.state.isPanelStateDefault() || !this.state.isPanelStateValid()) {
      this.setDefaultPanelState()
    }
    // If panel should be opened according to the state, open it
    if (this.state.isPanelOpen()) {
      if (this.uic.hasModule('panel')) { this.api.ui.openPanel(true) } // Force close the panel
    }
```
(see https://github.com/alpheios-project/alpheios-core/blob/master/packages/components/src/lib/controllers/app-controller.js#L997-L1003)

and as `setDefaultPanelState()` is defined as
```
  setDefaultPanelState () {
    if (!this.uic.hasModule('panel')) { return this }
    this.state.setPanelClosed()
    return this
  }
```
Then, combined, it will be:
```
    if (this.state.isPanelStateDefault() || !this.state.isPanelStateValid()) {
      if (!this.uic.hasModule('panel')) { return this } // I.e. do nothing
      this.state.setPanelClosed()
    }
    // If panel should be opened according to the state, open it
    if (this.state.isPanelOpen()) {
      if (this.uic.hasModule('panel')) { this.api.ui.openPanel(true) } // Force close the panel
    }
```
In both cases nothing is done if there is no `panel` module available. We open a panel if its state is `open` and close if the state is either `default` or invalid. In all other cases we do nothing. So I think it must be pretty much what we had before.

The complicated part is that the `state` is provided from an outside of an app: both webextension and embed lib have their own state object. That bothers me because an app controller cannot know what possible states will it be dealing with. Because of this we, in the app controller, cannot expect anything about the states being provided. I think this is no good.

That's one of the reasons I was suggesting to simplify state management. I as thinking about using Vuex but I agree it's not a very good idea because of a dependency that it will create. But I still think something can and needs to be done here. Maybe we can do the change when we'll convert to code to adapt to the Manifest V3, because it's a webextension that using the state the most?

What do you think about that? Can you suggest any improvements to the current piece of code?

balmas: I think the current code is fine for now.

I agree that we need to deal with the co-managed state (client app vs app controller) better. Not sure when or how is best to do that right not, but for the moment waiting until we adapt for Manifest 3 seems as good of an idea as any!

kirlat:
@irina060981, thanks for you thought and comments! These are good ideas and suggestions, and I will definitely need to think about it for a while. We could discuss them during the call tomorrow.

PR #487, opened 2020-07-03
L10n, getSelectedText improvements

kirlat:
`L10n` was used in several packages, and some packages kept their own version of the library. I've moved the `L10n` to the `data-models` and modified packages so that they will use one version of `L10n` form the `data-models`. 

I've also mode some changes to simplify handling of `get selected text` request.

Slightly rewrote code that is using `overrideHelp` variable to make it clearer. I've also added detailed comments explaining the purpose of this variable.


packages/components/src/lib/controllers/selection-controller.js
```diff
@@ -1,12 +1,25 @@
...
 /**
  * `SelectionController` manages all aspects of text selection.
  */
 export default class SelectionController {
-  constructor () {
+  /**
+   * @param {Function} getDefaultLangCodeFn - A function that returns a default language code in ISO 639-3 format.
+   *        This language code will be used to determine a language if language info is not provided by the page.
+   */
+  constructor (getDefaultLangCodeFn = () => 'lat') {
```
irina060981, line 13: I didn't quitely understand, why do you use a function for defining default languageCode.
Default languageCode should be a constant by its sense.
I din't find in the code where it is used as function.

packages/data-models/src/driver.js
```diff
@@ -36,6 +36,9 @@ import WordList from './word-list.js'
 import TreebankDataItem from './treebank_data_item.js'
 import Logger from './logging/logger.js'
 
+import L10n from './l10n/l10n.js'
```
irina060981, line 39: From my point of view it would be better to create a separate l10n package,
and the locale file (with translations) should stay inside the package they belong to.
I see the following reasons for that:
   - this way it could be used inside any alpheios application (even if it doesn't need the whole alpheios-core)
      for example may be it is useful to use it in a new alignment editor library, but I want to avoid to upload there the whole huge alpheios-core
    - we could store localization files withing the packages to simplify there handling and navigation
    - l10n has no dependencies for other packages (including data-modules) so it could became really a small and light package

packages/components/src/lib/controllers/selection-controller.js
kirlat: Default language code is calculated via options as `return this.options.overridePreferredLanguage ? this.options.textLangCode : this.featureOptions.items.preferredLanguage.currentValue`. Because of this, it can change any time during the project lifetime. So we need it to be calculated dynamically, It can either be a getter or a function. I've decided to use a function because it serves as a good indication of a dynamic nature of the value.

packages/data-models/src/driver.js
kirlat: I agree that it's better to have `L10n` as a separate package. @balmas, what do you think?

balmas: yes that seems reasonable.

PR #488, opened 2020-07-07
L10n split

kirlat:
Moved L10n into a separate package, optimized build configurations to eliminate duplicated code and reduce output sizes. Updated dependencies.


balmas:
> @kirlat, I think that th separation is good.
> But I thought that it would be completely a separate package and this way it would be able to use inside other applications without alpheios-core.
> So we continue to use it only with alpheios-core?

I asked to keep it in alpheios-core for simplicity. If it turns out we need it outside of alpheios-core, I think we should publish it as an npm package, like we do with components.

kirlat:
> But I thought that it would be completely a separate package and this way it would be able to use inside other applications without alpheios-core.
> So we continue to use it only with alpheios-core?

That's what [Babel](https://github.com/babel/babel/tree/main/packages), [Jest](https://github.com/facebook/jest/tree/master/packages), and others do. Packages are grouped together in one monorepo yet we can install each individual package separately. It's much easier to manage packages when they're in one place.

PR #494, opened 2020-07-23
Tests of the UI Controller

kirlat:
Created tests for the UI controller. Made multiple optimizations to the UI controller and surrounding classes.

Will also submit corresponding PR for embedded lib because some changes are required there too.


packages/components/tests/helpclasses/base-test-help.js
```diff
@@ -41,223 +41,278 @@ export default class BaseTestHelp {
...
   }
 
-  static baseVuexStore () {
-    return new Vuex.Store({
-      modules: {
-        popup: {
-          namespaced: true,
-          state: {
-            visible: false
+  /**
```
balmas, line 44: I don't suppose there is any way to automatically generate the mock for the Vuex store from the code itself?

kirlat: I was following the practices currently established in the `BaseTestHelp` class, but I'm wondering if real Vuex objects, not mocks, would be nearly as performant. I probably can run a series of tests to see how would it be if we'll use real life objects instead of the mocks. If real life objects be simple, there will be no serious performance impact. Making them simple will benefit the clarity of the code as well. So if the code is too heavy and complex to be used in tests maybe we should simplify and split it until it becomes lightweight enough? What if we try to make real business object mockable (not sure we can but we can try, we'll need to make things that are not suitable for tests easily replaceable during test time as we did with the `z-index` method) instead of creating mocks for them?

kirlat:
@irina060981, I've merged this so that @monzug could test it, but please leave your comments here anyway. I will make the necessary amends and merge them on top of what's in there now. Thanks!

packages/components/tests/helpclasses/base-test-help.js
balmas: I think we need to also ask ourselves what really needs to be mocked and whether the fact that we have this need means we still have too tightly coupled code.

irina060981:
I thought before (when I ws creating BaseTestHelp class) how we could do it easier to use Vuex Store in tests, and I saw the following ways:

1. we could create unified mocks - and as we are going further - it would become more and more complex,
    and I believe that coments here tell that we reached some complexity.
    but I started to go this way - because on the first steps it allowed to reduce current complexity of unit tests creating process

2. and we could reduce dependency inside components - I think that this is more perspective way, because it would made code more stable. Why do we need o rich mocks:

- we couldn't even create or initialize vue-components without rich defined Vuex input,
- we need to execute a chain of getters to come to some application state (and this way we also need to define and know what exactly should be defined to create chains)
- we have strong dependency on background variables instead of heaving better dependency on input methods variables

We could try to duplicate the real Vuex - it is not really difficult, I believe - we need simply go through real steps for each unut test, but I think that contradicts unit tests philosofy and would create a big complexity (that's why I created a simpler BaseTestHelp)

But I think that it is worth to reduce complexity of the code and reduce cross-dependencies.
And it will make easier to use our application more flexible in future.

irina060981:
I look through tests - @kirlat , I think that you have already done a good job for creaing tests for UIController.
I have no notices to them - they are really well structured and realy clear.

I have only one question - is this only the first step of covering "old UIController" with tests? I believe these tests covers near 1/3 of the functionality of the "old UIController". I believe other functionality are in other controllers now.
Do we have a task to cover with tests all controllers?

kirlat:
> I have only one question - is this only the first step of covering "old UIController" with tests? I believe these tests covers near 1/3 of the functionality of the "old UIController". I believe other functionality are in other controllers now.
> Do we have a task to cover with tests all controllers?

Yes, I'm planing to add tests for other functionality from other controllers as well. You're right that the "new" UI Controller is just a part of what the "old" UI controller was.

balmas:
>But I think that it is worth to reduce complexity of the code and reduce cross-dependencies.
And it will make easier to use our application more flexible in future.

Yes, I agree with this. I would rather we put our time into reducing complexity and cross-dependencies than focusing specifically on simplifying the process of creating mocks. That is, I think if every time we find a need  to create a mock with methods or properties that duplicate what is the code, we will have identified a tight coupling that we should question.

kirlat:
@balmas, @irina060981: What if we use the following approaches to testing?

Try to avoid using mocks as much as possible. Make "real" objects more test-friendly. Reduce dependency between objects. If mocking is unavoidable, mock not the whole object, but just a part of it (i.e. a method).

We could use the following principles to achieve this:
1. Try to straighten links between object to make them less dependent on each other (I've made some modifications of the kind within the UI controller to achieve this and make testing easier).
2. Try to write a "defensive" code. If thee is an external dependency, assume that it may not be valid and provide reasonable defaults that may make the component functional regardless. This is probably mostly important for the Vue components. If something refers to `moduleX.itemA` assume that both may not exist. This would probably require instead of referring to `moduleX.itemA` directly in the template create a computed prop. Inside a computed prop we can check if `moduleX.itemA` exists and if not, return a default value. Computed props may make the testing easier: instead of examining an HTML template (which might be slow in Jest), we can check a return value of computed props. We might also be probably able to mock individual computed props if necessary.
3. Modules can, along with its class, export the Vuex settings object that is used to create Vuex modules. With this, we can import Vuex settings in the test environment and create Vuex store modules without creating heavier module object (i.e. PopupModule, PanelModule, etc. being the heavier object vs. the Vuex settings) itself.
4. We should think about exposing public API of modules in a way similar to exposing Vuex settings objects. That will allow for other objects to use an API without creating a module itself.
5. Reorganize methods to make objects more testing friendly. Create methods that can be mocked easily, if necessary. It's better to mock a simple method than the whole object.

What do you think about the approaches described above?

balmas:
these all sound like good practices to me.

irina060981:
I like this approach too!

Issue #478, opened 2020-06-24
Remove Dual and Plural from Greek noun tables


monzug:
Dual and Plural should not be repeated in the Greek noun tables. we should keep only one dual for nominative, accusative, vocative and one plural for nominative, vocative.
![Screen Shot 2020-06-19 at 3 32 48 PM](https://user-images.githubusercontent.com/41396793/85582443-ba1dd900-b63d-11ea-9172-f44b822eff33.png)

balmas:
Alpheios Reading Tools 3.3.1 build qa.20200803307
Alpheios Components 3.3.0-qa.20200803296

monzug:
this has been fixed in  Alpheios Reading Tools 3.3.1 build null.20200803479
Alpheios Components 3.3.0-qa.20200803296

PR #498, opened 2020-07-31
I478 28 inflections

### LINKED ISSUE #28: Comparative and superlative adjective table and Whitaker
an example of comparative adjective table could be found at 
http://dcc.dickinson.edu/sites/all/files/comparative_adjectives.pdf
In some latin grammars, the comparative adjective table is identical to the 3rd noun declension: melior, melior-is, melior-i etc.
problem is that Whitaker has different endings: meli-or, meli-oris, meli-ori etc.
If we add a comparative table with the endings as above pdf, we will still not have matches.
I think that Allen and Greenough is also using the endings as per the above pdf.
this also applies for the superlative . 
How do we want to resolve this?

from Allen and Greenough: chapters 122-127

  | melior , better STEM meliōr- for meliōs- |   | plūs , more STEM plūr- for plūs-
-- | -- | -- | --
  |   | SINGULAR
  | M., F. | N. | M., F. | N.
NOM. | melior | melius | ----- | plūs
GEN. | meliōris | meliōris | ----- | plūris
DAT. | meliōrī | meliōrī | ----- | -----
ACC. | meliōrem | melius | ----- | plūs
ABL. | meliōre ( -ī) | meliōre ( -ī) | ----- | plūre
  |   | PLURAL
NOM. | meliōrēs | meliōra | plūrēs | plūra
GEN. | meliōrum | meliōrum | plūrium | plūrium
DAT. | meliōribus | meliōribus | plūribus | plūribus
ACC. | meliōrēs ( -īs) | meliōra | plūrēs (-īs) | plūra
ABL. | meliōribus | meliōribus | plū ibus | plūribus

balmas:
Fixes #28 and #478


balmas:
> Looks well!
> @balmas , should I update `inflection-input`
> https://alpheios-misc-dev.s3.us-east-2.amazonaws.com/inflection-input/index.html
> ?

Hmm. Probably not a bad idea, although this only affects Latin which isn't in that tool right now.

Issue #484, opened 2020-06-30
ability to use Treebank data for disamibugation only


balmas:
requested by @nevenjovanovic

a project I did with students included using Perseids only to disambiguate
parses and lemmatizations, without creating treebanks. Is there a
possibility to say to Alpheios that I have disambiguations (in the ALDT
format), but no treebanks? That is, to show just the disambiguation
symbol, but not the diagram symbol?

This could be pedagogically useful because at the early learning stage it
helps to have a disambiguated (grammatically commented) text...

balmas:
this should be doable. it might require a new API call to Arethusa to query whether or not any relations have been defined for a sentence.

balmas:
this can be tested in

Alpheios Embedded Library 3.3.0-qa.20200807266
Alpheios Components 3.3.0-qa.20200807247 

At https://alpheios-misc-dev.s3.us-east-2.amazonaws.com/treebank-test-page/test.html

In the user-report tests, with sentence Τοὺς οὖν πρώτους τῶν ἀνθρώπων μηδενὸς τῶν πρὸς βίον χρησίμων εὑρημένου ἐπιπόνως διάγειν , γυμνοὺς μὲν ἐσθῆτος ὄντας , οἰκήσεως δὲ καὶ πυρὸς ἀήθεις , τροφῆς δ’ ἡμέρου παντελῶς ἀνεννοήτους .

Any word in this sentence should produce a disambiguation icon in the popup. The tree icon will be next to the disambiguation error still, but the tooltip has been clarified to say "Disambiguated with treebank data" and there is no longer a toolbar button to produce the tree in popup header nor in the toolbar.

monzug:
I see the tooltip with disambiguation with trrebank data. the toolbar button is gone but I do not have any other way to open the diagram. how am I supposed to open the diagram?
tested in chrome and FF on PC, build   Alpheios Embedded Library 3.3.0-qa.20200807266
Alpheios Components 3.3.0-qa.20200807377

balmas:
>how am I supposed to open the diagram if tree display is disabled using data-alpheios_tb_morph_only=true?

That's the point -- the diagrams are disabled because in fact in this data set no relationships between words have been annotated only the morphology. So the diagram would be empty if you opened it.

monzug:
ok. I was rewriting the sentence to check if this was the point. Thanks

PR #503, opened 2020-08-05
I484 suppress tree

balmas:
This provides a solution for #484 by adding support for a new treebank data instruction, `data-alpheios_tb_morph_only` which if set (to anything except the string 'false') will suppress display of the treebank icon to show the tree diagram, but still uses treebank data for disamibugating the morphology. It's a bit of a hack, but so is using treebank data in this way. It fulfills the user request until the lexical query refactoring and additional morph data sources are available.

It also fixes a bug in the lexis.js that I think was introduced in the ui controller refactoring -- previously we didn't crash if the config service was unavailable, but that was happening now, I think due to changes in the way the lexisConfig object is created. I think this code is more correct anyway, because it's not just if the config object is defined that cedict is available. It is possible there could be a config object without cedict settings in which case the cedict service should also be disabled.


packages/data-models/src/treebank_data_item.js
```diff
@@ -18,7 +18,7 @@ export default class TreebankDataItem {
     this.wordIds = []
     this.sentenceId = null
     this.doc = null
-
+    this.suppressTree = false
```
kirlat, line 21: Maybe to add a comment explaining why do we suppress a tree (just to be able to recall it later)?

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -483,12 +492,12 @@ Lexis.api = (moduleInstance, store) => {
     },
 
     loadCedictData: async () => {
-      if (!moduleInstance._lexisConfig) { return } // Do nothing if Lexis configuration is not available
+      if (!moduleInstance.hasCedict()) { return } // Do nothing if Lexis configuration is not available
```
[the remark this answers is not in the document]
balmas: yes will fix.

packages/data-models/src/treebank_data_item.js
balmas: good idea :-) will fix.

Issue #514, opened 2020-08-11
Missing ending in the superlative table for some irregular adjectives


monzug:
Irregular adjectives, such as optimus, pessimus, dextimus, citimus, extremus, supremus, do not have a match in the superlative adjective table.
Should I add the ending mus, ma, mum to the table? in Whitaker, we have the mus, ma, mum endings but in Bennett the ending is imus. Need to check more grammars.
related to issue #28

balmas:
I'll leave this one up to you @monzug

monzug:
miss the Spilimbergo library!!!!!

monzug:
after checking few grammars and considering that  "The comparative and superlative are really new stems, and are not strictly to be regarded as forms of inflection." I decided that by adding the  "mus" ending with a footnote, Bennett would not mind...

monzug:
add footnotes 4 and endings in the superlative

balmas:
Components qa.20200812380

monzug:
footnotes 4 and endings in ma, mus, mum are in the superlative table

PR #515, opened 2020-08-11
I28 fixes

balmas:
Fixes remaining issues in #28 and #514


balmas:
> Looks well!
> 
> @balmas , may be in future we should think about refactoring createTable method in views to reduce duplicate code and increase readability.
> We could do this with passing features names the follwoing way:
> add a method `constructTable ` with arguments:
> { tableFeatures, columsFeatures, rowsFeatures, rowTitles, fullWidthRowTitles}

Thats a good idea. Will make this a new issue for future refactoring.

PR #516, opened 2020-08-12
App controller's test

kirlat:
This build adds tests for the App Controller. The problem that I've reported earlier that seems to be Vuex related was, as I figured out finaly, a subject to a different thing instead: functions that are listening to pub/sub events. Some of those listeners might be called when the object environment (the app controller, the Vuex store, and the related objects) are already destroyed after the test is finished. 

This does not happen when tests are run individually as, I think, they runs faster and listeners are called when objects still exist. When tests are run in sequence. some parallel execution happens, and calls to listeners are delayed. As a results, listeners might be called when the environment that they are supposed to be in does not exist any more.

To solve this, I've created a test version of a `create()` method that omits registration of those listeners that are known to be affected by those problems. That allowed multiple tests to be executed without any problems.


packages/components/tests/lib/controllers/selection-controller.test.js
```diff
@@ -0,0 +1,175 @@
...
+
+  afterEach(() => {
+    jest.resetModules()
+  })
+
+  afterAll(() => {
+    jest.clearAllMocks()
+  })
+
+  it('UIController - constructor: should create an instance with default arguments', () => {
```
irina060981, line 17: I belive, that titles should start from SelectionController.
I defined the title format this way - to be able easily to identify what test case is failed, when a bunch of tests are executed.

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    mouseMoveLimitedById: 'docs-editor-container'
+  }
+
+  /*
+  Since events are asynchronous, their listeners may be called in a test environment
+  after the application is tear down. To prevent errors related to this, a special
+  version of the create() method is created for the test environment. In it, event listeners
+  that may have issues like described above are not registered.
+   */
+  AppController.jestCreate = (state, options) => {
```
irina060981, line 37: it is an interesting approach!

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    let appController = new AppController(state, options) // eslint-disable-line prefer-const
```
balmas, line 38: Instead of having a test-specific create method couldn't we move the subscription calls to its own method in AppController with a parameter which indicates whether or not to skip the query subscriptions?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
-    expect(uiC.options.template).toHaveProperty('draggable')
-    expect(uiC.options.template).toHaveProperty('resizable')
+  it('AppController - create: an instance created should have L10n and Lexis modules', () => {
+    appC = AppController.jestCreate(uiState)
+    expect(appC.hasModule('l10n')).toBeTruthy()
+    expect(appC.hasModule('lexis')).toBeTruthy()
+  })
 
-    expect(uiC.zIndex).toBeGreaterThan(0)
+  it('AppController - setOptions: should combine options provided and the default ones', () => {
```
balmas, line 133: this test doesn't look any different than the previous one that tests options. Is it actually testing if options are combined?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    await appC.init()
+    appC._platform = { isMobile: true, isDektop: false }
+    const [event, eventParams] = appC.textSelectorParams
+    expect(event.name).toBe('LongTap')
+    expect(eventParams).toBeUndefined()
+  })
 
-    uiC.addImportantMessage('fooImportant1')
-    expect(uiC.panel.panelData.messages).toEqual([ 'foomessage1', 'foomessage2', 'fooImportant1' ])
+  it('AppController - textSelectorParams: should return correct parameters no trigger is specified on mobile', async () => {
```
balmas, line 356: this test looks identical to the previous -- I think the create should be taking instead an empty object for options?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    expect(event.name).toBe('MouseMove')
+    expect(eventParams).toEqual(defaultMousemoveParams)
+  })
 
-    expect(uiC.popup.vi.popupData.notification.visible).toBeTruthy()
-    expect(uiC.popup.vi.popupData.notification.important).toBeTruthy()
-    expect(uiC.popup.vi.popupData.notification.showLanguageSwitcher).toBeFalsy()
-    expect(uiC.popup.vi.popupData.notification.text).toEqual('fooImportant1')
+  it('AppController - textSelectorParams: should return correct parameters for the "dblClick" trigger on desktop', async () => {
+    appC = AppController.jestCreate(uiState, { textQueryTriggerMobile: 'dblClick' })
```
balmas, line 389: textQueryTriggerMobile should be textQueryTriggerDesktop?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    appC._platform = { isMobile: false, isDektop: true }
+    const [event, eventParams] = appC.textSelectorParams
+    expect(event.name).toBe('MouseDblClick')
+    expect(eventParams).toBeUndefined()
   })
 
-  it('6 AppController - showLanguageInfo methods', () => {
-    uiC.showLanguageInfo()
-    let languageName = AppController.getLanguageName(LMF.getLanguageIdFromCode(uiC.panel.options.items.preferredLanguage.currentValue)).name
+  it('AppController - textSelectorParams: should return correct parameters for the "dblclick" trigger on desktop', async () => {
```
balmas, line 397: textQueryTriggerMobile should be textQueryTriggerDesktop?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    expect(event.name).toBe('MouseDblClick')
+    expect(eventParams).toBeUndefined()
+  })
 
-    expect(uiC.panel.panelData.notification.visible).toBeTruthy()
-    expect(uiC.panel.panelData.notification.important).toBeTruthy()
-    expect(uiC.panel.panelData.notification.showLanguageSwitcher).toBeTruthy()
-    expect(uiC.panel.panelData.notification.text).toEqual(l10n.messages.TEXT_NOTICE_CHANGE_LANGUAGE.get(languageName))
+  it('AppController - textSelectorParams: should return correct parameters for the null trigger on desktop', async () => {
+    appC = AppController.jestCreate(uiState, { textQueryTriggerMobile: null })
```
balmas, line 407: textQueryTriggerMobile should be textQueryTriggerDesktop?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    const [event, eventParams] = appC.textSelectorParams
+    expect(event.name).toBe('MouseDblClick')
+    expect(eventParams).toBeUndefined()
+  })
 
-    expect(uiC.popup.vi.popupData.notification.visible).toBeTruthy()
-    expect(uiC.popup.vi.popupData.notification.important).toBeTruthy()
-    expect(uiC.popup.vi.popupData.notification.showLanguageSwitcher).toBeTruthy()
-    expect(uiC.popup.vi.popupData.notification.text).toEqual(l10n.messages.TEXT_NOTICE_CHANGE_LANGUAGE.get(languageName))
+  it('AppController - textSelectorParams: should return correct parameters no trigger is specified on desktop', async () => {
```
balmas, line 415: textQueryTriggerMobile should be textQueryTriggerDesktop?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
+    const activateSelectorSpy = jest.spyOn(appC._selc, 'activateSelector')
+    appC.activateTextSelector()
+    expect(activateSelectorSpy).toBeCalledTimes(1)
   })
 
-  it('7 AppController - showStatusInfo methods', () => {
-    uiC.showStatusInfo('fooSelectionText', latID)
-    expect(uiC.panel.panelData.status.languageName).toEqual('Latin')
-    expect(uiC.panel.panelData.status.selectedText).toEqual('fooSelectionText')
+  it('AppController - activateTextSelector: should call activateSelector on the selection controller', async () => {
```
balmas, line 453: duplicate of prior test?

packages/components/tests/lib/controllers/app-controller.test.js
```diff
@@ -1,839 +1,1389 @@
...
-    expect(uiC.panel.panelData.messages).toEqual([])
+  it('AppController - handleEscapeKey: should do nothing if the UI is inactive', async () => {
+    const event = {}
+    const nativeEvent = {
+      keyCode: 27
+    }
+    const uiState = BaseTestHelp.createUIState({ isActive: () => false })
+    appC = AppController.jestCreate(uiState)
+    await appC.init()
+    await appC.activate()
```
balmas, line 1016: I'm confused about why this test works.  Shouldn't appC.activate() cause the uiState to be active?

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: The purpose of this test is to check that the options supplied would be merged properly with the default options. So it creates a custom options object and checks whether the `AppController.setOptions()` will place it into the resulting output.

We can also add a check that without the `customOptions` object supplied the `AppController.setOptions()` output will not contain the custom options object.

Does it make sense?

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: Oh, yes, I've overlooked that. Thanks for catching! Will fix.

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: Yes, will fix that.

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: Will fix

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: Will fix that too

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: Yes, definitely

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: `create()` should be called here without any parameter, will update

packages/components/tests/lib/controllers/app-controller.test.js
kirlat: > Shouldn't appC.activate() cause the uiState to be active?

In the real objects, yes. `appC.activate()` would call `activate()` of the UI controller. We, however, use the mock of the `uiState` and the `activate()` method, when called upon it, will do nothing. Even though if it would be the real object, we redefine its `isActive` method so that it will always return false: `const uiState = BaseTestHelp.createUIState({ isActive: () => false })`. Because of that, the UI would be identified as inactive.

I agree it's a hacky way to test, but I think it's probably OK considering that the UI state is a mock, not a fully-functional object. I think we should test with the real version of the `uiState` ideally, but considering complexities of that and the fact that we're planning to change the way options are stored and handled maybe we can postpone that for later?

I can add a comment explaining the situation.

What do you think?

packages/components/tests/lib/controllers/app-controller.test.js
balmas: It makes sense, I just wasn't seeing how this test differed from the one on lines 110-119.  I.e. it didn't seem to be testing a merge of customOptions with default options, it just tested that what was supplied is what was set.

packages/components/tests/lib/controllers/app-controller.test.js
balmas: sure, a comment would be helpful!

packages/components/tests/lib/controllers/selection-controller.test.js
kirlat: Oh, I've overlooked that completely. Thanks for catching this! Will rename.

PR #527, opened 2020-08-13
Githubactions

balmas:
this moves the build, tagging and publication of the qa build of components to a github action.

previously, we ran tagged-commit manually, pushed the branch and the tag to qa and then travis published the npm package and the release.

with this change, the qa build, tagging and publication of the release and the npm package are all done by the GitHub job.

Next steps will be to add the production and ftqa branches to the workflow.


github-build.mjs
```diff
@@ -0,0 +1,47 @@
...
+    console.error('Build process failed:', error)
+    process.exit(2)
+  }
+  console.log('Rebuilding of a components library has been completed')
+  try {
+    console.info(core)
+    core.default.setOutput('buildName',buildInfo.name)
+  } catch (error) {
+    console.error('Failed to set output variable:', error)
+    process.exit(1)
```
kirlat, line 45: Should we have a different exit code here? We've used the "1" already

balmas: oh that's a good suggestion. I'll change that.

PR #538, opened 2020-09-28
Dev wordlist refactoring

kirlat:
This is a first phase of the lexical query refactoring. It was tested for Latin and Greek mostly, and is not production ready by any means. It needs many improvements before we'll be able to use it fully. However, it shows all aspects of a redesigned architecture.

The purpose of the PR is to show the concept and offer if for discussion.

All the new files are placed into two directories: `app` and `data-model`. The former contain the files that would be used on the client side, by the app controller and other objects that belong to it. The latter contains the files responsible for data retrieval. They can theoretically be hosted on the server side (but that would require some code changes). These directory division is not final and may change in the future.

Many decisions in the code are dictated by separation of concerns. I've tried to add comments to explain design decisions, but I know it needs much more. I will add it in the next few days.

The way some code flows are implemented is temporary. It will be improved in the future.

The tests are almost non-existent, I will work on them in the next few days. I (and other people) has problems trying to make the `fetch` that is used by Apollo work in Jest environment. I was not able to solve it so far but hopefully I will. It, unfortunately, blocks testing of all objects that are Apollo-dependent.

Please let me know what do you think. Sorry for too many changes in one PR, but some things were impossible without changing the other and so on. One thing leads to another.


packages/components/src/lib/controllers/app-controller.js
```diff
@@ -377,7 +384,11 @@ export default class AppController {
       buildNumber: this._options.app.buildNumber,
       storageAdapter: this._options.storageAdapter
     })
-    // All options has been loaded and initialized after this point
+    // All options are initialized at this point
+
+    // We do have dependency on the DataModelController only because we're using local fields
+    // with resolvers from the DataModelController. This dependency will not exist with the remote GraphQL service.
```
balmas, line 390: Even when we have a remote GraphQL service I expect we will also have a client-side service as well (e.g. for treebank and other client-side data)

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1141,7 +1152,8 @@ export default class AppController {
         this._store.commit('ui/addMessage', this.api.l10n.getMsg('TEXT_NOTICE_LEXQUERY_COMPLETE'))
         break
       case LexicalQuery.resultStatus.FAILED:
-        this.showLanguageInfo(data.homonym)
+        // TODO: Can we do anything better if homonym is not available?
```
balmas, line 1155: in the future, this might be a place to prompt the user to query a word level dictionary resource if there is one (or some other linked word-level resource -- for example, it's a place we could do better right now, if we don't have a result for a homonym in Latin, we could still offer the option to search the usage examples)

balmas: similarly we might prompt the user to create their own analysis...

balmas (commented):
still making my way through this, just wanted to submit the few comments and questions I have so far so that they don't get lost.  Will pick it up again first thing tomorrow.

packages/components/src/app/word-query/word-query-adapter.js
```diff
@@ -0,0 +1,115 @@
...
+   * Similar to `observableQuery()`, but with its own callback implementation.
+   *
+   * @param {object} options - Data that will be passed to the GraphQL query.
+   * @param {object} options.variables - Variables to be passed to the GraphQL query.
+   * @param {number} options.pollInterval - An interval in milliseconds specifying how often a new data will be retrieved.
+   * @param {number} options.maxIterations - How many iterations to do before stopping the query.
+   * @returns {Promise<*>} - A promise that is resolved with GraphQL data or is rejected with an error.
+   */
+  async getFromObservableQuery ({ variables, pollInterval = 1000, maxIterations = 10 } = {}) {
+    return new Promise((resolve, reject) => {
```
irina060981, line 98: I think that it is pretty the same - define a method as async and return a new Promise, I believe that async is redundant.

And also I didn't finally understand the structure of this request:

we have `this.observableQuery`  that does the following steps:

(by the way I think that having the same names for a local variable and the whole method around of this variable is rather confusing and makes the code worse readable, also you don't even return the local variable as the result of this method)
 - define `observableQuery` as `watchQuery` of Appolo client that is equal to the `ObservableQuery`
 - define `subscribe` for this `observableQuery` and on received good results inits dataCallback with predefined data result (I believe you would add error handling here later)

and then here `getFromObservableQuery` - we redefine dataCallback with the different argumnets list?
Was this `data.word.homonyms, data.word.state, data.word.errors` and becomes this `(data, loading)` ?
Or may be I have mixed the same names for different intances?

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,314 @@
...
+    return path
+  }
+
+  // Runs all the jobs within the execution path.
+  static async _execute (executionPath, context) {
+    try {
+      for (const batch of executionPath) {
+        if (!context.cancelled) {
+          const jobs = batch.map(job => WordQuery[job](context))
+          await Promise.all(jobs)
```
irina060981, line 82: I didn't understand the construction here too.
what does this mean - `WordQuery[job](context)` without word `new`?

And also I didn't understand what is the heierarchy here and what would be do in sequence and what in parallel?
- job
- execution path
- batch

and when you pass `context` - to what object do you pass it - to the job or batch?
And if the batch is someting like - getMorph, disambiguate, getLexicons, the could contex be different for such steps? For example te result of previous step could add limits to the next steps?

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,314 @@
...
+    } catch (err) {
+      Logger.getInstance().error(err)
+    }
+    WordQueryResult.loadingStopped(context.result.state)
+    context.result.state = Object.assign({}, context.result.state)
+    return context
+  }
+
+  // Retrieves morphological data from Tufts
+  static async _getTuftsMorphology (context) {
```
irina060981, line 94: do these methods `_getTuftsMorphology`, `_disambiguate` and others should be considered as query steps? graphQl results in the execution path? 
If it is so I suggest to move it to a separate file , because it should be updated independently from the query construction process. what do you think?

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -312,7 +326,61 @@ export default class Lexis extends Module {
...
+    return wordIDs
+  }
+
+  async getWordQueryData ({ variables, pollInterval = 1000, maxIterations = 10, source, languageId } = {}) {
+    return new Promise((resolve, reject) => {
+      try {
+        this._adapters.wordQuery.observableQuery({
+          variables,
+          // This callback is called every time the GraphQL query is updated, until the query is complete.
+          dataCallback: (data, state, errors) => { // TODO: How the errors object should be handled?
```
irina060981, line 338: and here again we have another dataCallBack for the same wordQuery?

irina060981 (commented):
If to be honest I didn't finally understand this mix of Promises and redefined dataCallbacks, would be happy to read more explanations :) 
And also - would we have only one query with all steps similiar to the current LexicalQuery approach.
Or would it be several queries for different use cases?

packages/components/src/app/word-query/word-query-adapter.js
```diff
@@ -0,0 +1,115 @@
...
+              read: wordQueryResolver
+            }
+          }
+        }
+      }
+    })
+
+    this._gql = new ApolloClient({
+      cache,
+      link: new HttpLink({ uri: '/graphql', fetch }),
```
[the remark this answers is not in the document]
kirlat: Here `fetch` is a reference to the `fetch` API of the browser: https://www.apollographql.com/docs/link/links/http/#fetch-polyfill. But that creates an issue with testing in node/Jest environment. We need to use a polyfill. The one I've used, `cross-fetch` has some issues that I was not able to resolve so far.

The option to provide a fetch-compatible API instead of using the only one provided by Apollo is part of a flexible Apollo link architecture.

packages/components/src/app/word-query/word-query-adapter.js
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * An asynchronous GraphQL query. Data is returned when the query is fully completed.
+   *
+   * @param {object} options - Data that will be passed to the GraphQL query.
+   * @param {object} options.variables - Variables to be passed to the GraphQL query.
+   * @param {Function} options.dataCallback - A function to be called every time new data arrives.
+   * @param {number} options.pollInterval - An interval in milliseconds specifying how often a new data will be retrieved.
+   * @param {number} options.maxIterations - How many iterations to do before stopping the query.
+   */
+  observableQuery ({ variables, dataCallback, pollInterval = 1000, maxIterations = 10 } = {}) {
```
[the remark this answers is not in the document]
kirlat: What I've meant was that it is asynchronous by nature but it does not fit the JS `async` specification. The `observableQuery` is created for situations when not all query results became available immediately but may be delivered over time. In our case, for example, it can take a lot of time to retrieve all lexical data, but some results (like lexemes) may be delivered to the client before the query is fully complete (i.e. until short definitions are obtained).

The `observableQuery` is aware of the dynamic nature of the query and it calls the user-provided data callback function (`dataCallback`) every time when query results are updated. In other words, `observableQuery` knows that the query is "live" and it polls the query data every X milliseconds or until the maximum number of poll attempts is reached (those numbers needs careful tweaking for production).

Do you think there could be some other way to describe it in order to avoid confusion with the JS `async` specs?

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,82 @@
...
+      contextForward: variables.contextForward,
+      contextBackward: variables.contextBackward
+    })
+    // The words map will store references to the results stored in the Apollo cache
+    if (!storage.words) { storage.words = new Map() }
+    if (!storage.words.has(key)) {
+      const wqResultWatcher = {
+        // This function will track changes in the top-level props of the result object
+        set: function (obj, prop, value) {
+          let wordReactive = storage.words.get(key) // eslint-disable-line prefer-const
```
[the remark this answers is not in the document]
kirlat: This should not happen right now because the function is used after the key is set (https://github.com/alpheios-project/alpheios-core/blob/dev-wordlist-refactoring/packages/components/src/data-model/word-query/word-query-resolver.js#L63-L65), but I agree it is not reliable enough to rely on that. Things may change in the future. I'll add the guard.

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,82 @@
...
+          } else if (prop === 'state') {
+            // Update the value of the reactive variable
+            const state = WordQueryResult.stateToJsonObject(value)
+            // Need to recreate an object on order for reactive var to be updated
+            wordReactive({ ...word, state })
+          } else if (prop === 'errors') {
+            // Update the value of the reactive variable
+            const errors = WordQueryResult.errorsToJsonObject(value)
+            // Need to recreate an object on order for reactive var to be updated
+            wordReactive({ ...word, errors })
```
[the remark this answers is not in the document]
kirlat: The reason for the difference is that in case of a `homonymGroup` prop update the `value` is the "real" `HomonymGroup` object with methods while `state` and `errors` are JSON object representations with no methods on them. I think it's a good suggestion to use static methods in all the cases and I will update the code accordingly.

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,314 @@
...
+    if (queryParams.getLexemes) {
+      if (queryParams.useMorphService) {
+        getLexemesBatch.push('_getTuftsMorphology')
+      }
+      if (queryParams.useTreebankData) {
+        getLexemesBatch.push('_getTreebankData')
+      }
+      if (queryParams.useWordAsLexeme) {
+        getLexemesBatch.push('_getWordAsLexeme')
+      }
```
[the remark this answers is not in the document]
kirlat: The group of operations such as the retrieval of short definitions is called a "job". Jobs are grouped into batches. Jobs can run in parallel within the batch, and batches can run only one after the other.

Here is the executor code that provides an execution in that order (please let me know if it does not work as desired): https://github.com/alpheios-project/alpheios-core/blob/dev-wordlist-refactoring/packages/components/src/data-model/word-query/word-query.js#L79-L84

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,314 @@
...
+        if (!context.cancelled) {
+          const jobs = batch.map(job => WordQuery[job](context))
+          await Promise.all(jobs)
+        }
+      }
+    } catch (err) {
+      Logger.getInstance().error(err)
+    }
+    WordQueryResult.loadingStopped(context.result.state)
+    context.result.state = Object.assign({}, context.result.state)
```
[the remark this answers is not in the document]
kirlat: Yes, exactly. The proxy attached to the result object (`context.result`) tracks only the top-level props of the object: `context.result.homonymGroup`, `context.result.state`, and `context.result.errors`. If we assign to something below that level, as `context.result.state.loading = false`, the change will be not noticed.

So the key here is not to create a copy of an object, but rather to assign to the top-level prop of the `context.result`. The assignment as the one referenced above is the way I came up with that makes sense from code point of view. The other alternative that, I think, I've used somewhere else is a self-assignment: `context.result.state = context.result.state`. This works (the change is tracked) but it's a practice that can be frowned upon and not understood easily.

Actually, I think there is a problem with this code above because `Object.assign()` does a shallow copy only, and here we need a deep one. So I think I need to fix that.

Another way to create a copy is using `JSON.parse(JSON.stringify(context.result.state))`. But that is slower and looks probably more obscure.

So what would be the best way to do an assignment to `context.result.state`? Use `context.result.state = context.result.state` and explain in the comment why is it so? Or use a JSON method? Are there any other solutions? What do you think?

packages/components/src/lib/controllers/app-controller.js
kirlat: Yes, that's right, I'll change the comment to reflect that.

packages/components/src/app/word-query/word-query-adapter.js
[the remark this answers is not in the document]
balmas: I think this says it very well :-) " observableQuery knows that the query is "live" and it polls the query data every X milliseconds or until the maximum number of poll attempts is reached"

packages/components/src/data-model/word-query/word-query.js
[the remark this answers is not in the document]
balmas: ok thanks, that's clear.

packages/components/src/data-model/word-query/word-query.js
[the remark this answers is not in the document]
balmas: what do you think about adding a method that does just the same thing, but is named accordingly and then call that whenever we need to make a deep copy for reactivity? that would make the code clearer.

E.g. `WordQueryRequest.ensureReactivity(object)`

packages/components/src/lib/controllers/app-controller.js
kirlat: I think this would be a good extension of the functionality. I'll add a comment about that to the TODO block.

packages/components/src/app/word-query/word-query-adapter.js
```diff
@@ -0,0 +1,115 @@
...
+          fields: {
+            word: {
+              read: wordQueryResolver
+            }
+          }
+        }
+      }
+    })
+
+    this._gql = new ApolloClient({
```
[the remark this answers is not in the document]
balmas: this is a good point

packages/components/src/app/word-query/word-query-adapter.js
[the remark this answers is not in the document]
kirlat: Great, will put that into description.

packages/components/src/data-model/word-query/word-query.js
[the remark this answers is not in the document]
kirlat: That's a great idea! Will do that

packages/components/src/app/word-query/word-query-adapter.js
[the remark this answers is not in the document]
kirlat: I thought that `_qql.query()` would be nicer to read but I agree that it is confusing. Would it be OK to rename it to the `_apolloClient` (`_` is a convention to indicate that it is private as we don't want to implement Apollo client directly to the outside world)?

balmas (commented):
A few more comments on the code, plus a bigger question about the architecture.  I had to draw it out to try to understand it:

lexquerygraphqal

One thing I'm having trouble discerning is the line between the graphql client and the graphql "server". I recognize we don't have an actual graphql server in this instance, but a client-side implementation of the graphql server api. Or at least, that's my understanding.

As I understand it, the resolver of a query should be behind the graphql API. But it looks to me like we use this resolver as the read function for the InMemoryCache.  And it's not clear to me where the query execution actually happens  

We specify an endpoint of "/graphql" when we create the ApolloClient but that doesn't correspond to any real endpoint. So it looks like it's handing it off to the cache to do the actual resolution of the query.

I would expect the cache to have a reference to an adapter method to convert the JSON of the results of the query resolution (or execution) into the HomonymGroup object,  but instead it seems to responsible for actually resolving the query, which seems a little backwards.

What am I misunderstanding?

packages/components/src/app/word-query/word-query-adapter.js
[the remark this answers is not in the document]
balmas: that's okay for me

packages/components/src/data-model/word-query/word-query.js
kirlat: The job is a set of self sufficient operations such as retrieval of short definitions or of data from the Tufts adapter. Jobs are represented by their own static methods of the `WordQuery` class such as `_getTuftsMorphology()`, `_getTreebankData()`, `_disambiguate()`, `_getShortDefs()` and so on.

Jobs are grouped into batches. All jobs within one batch are executed in parallel.

A set of batches comprises the execution path. All batches are executed strictly one after the other. 

For example, if there is a word for which we need to retrieve data from the Tufts adapter and from the treebank, and get short definitions in addition to this then the execution path would be:
`[['_getTuftsMorphology', '_getTreebankData']['_disambiguate']['_getShortDefs']`.
It consists of three batches (top level array elements). The first batch consist of two jobs, the second and the third - of one. The content of the execution path arrays are the names of the methods of the `WordQuery` object.

The execution sequence for the execution path shown above would be:
1. `_getTuftsMorphology()` and `_getTreebankData()`, executed in parallel.
2. `_disambiguate()` - executed after the first two.
3. `_getShortDefs()` - executed the last.

`WordQuery[job]` refers to the static method of a WordQuery class (such as `WordQuery._disambiguate`). `WordQuery[job](context)` calls the method passing the `context` as a parameter. Because those methods are async, they return a promise. Those promises are gathered in the `jobs` array. `Promise.all(jobs)` converts multiple promises of a job into a single promise and in `await Promise.all(jobs)` we wait for all promises to resolve.

`context` is a shared state the is passed to each of the job methods. Each job (i.e. a static `WordQuery` method that matches the job) adds its own sub-object to the `context`. This sub-object contains the results of the job execution along with the status. For example, the `_getTuftsMorphology()` job adds the following to the context:
```
    context.tuftsMorphology = {
      state: {
        loading,
        available,
        failed
      },
      data: homonymGroup
    }
```
The `state` object contains results of the execution and the `data` prop contains the resulting `HomonymGroup` object.

Other jobs may use information from the previous jobs stored within the context. For example, a `_disambiguate` job may use Tufts and treebank adapter data from the context to produce a disambiguated homonym. It, for example, checks the `state.failed` field, and if it is truthy, then the results from the `data` field are not included into the disambiguation.

I will add a description of how it works to the documentation of the `WordQuery` class. Do you have any suggestions about how to make it clearer or how to simplify the workflow?

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,82 @@
...
+ * Creates an instance of WordQuery for the query resolution.
+ */
+export default class WordQueryResolver {
+  constructor ({ getLexiconsFn, getShortLexiconsFn } = {}) {
+    this._getLexiconsFn = getLexiconsFn
+    this._getShortLexiconsFn = getShortLexiconsFn
+  }
+
+  // This function will be called by Apollo when new GraphQL query will arrive.
+  resolve (_, { storage, variables }) {
```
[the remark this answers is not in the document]
balmas: I think if it's a convention of the graphql implementation we are using (i.e. apollo) then we should follow it but we could add some comments to our code to explain it.

packages/components/src/data-model/word-query/word-query.js
kirlat: Yes, those methods are the query steps. I think splitting them into a separate file is a very good idea because it would increase a separation of concerns. @balmas, would you agree with that?

packages/components/src/vue/vuex-modules/data/lexis.js
kirlat: Yes, we're using a custom `observableQuery` implementation here, but you're right that we probably don't need it. We can use the standard `observableQuery` implementation with the custom callback function. I will update the code to fix that. Thank you for a very good catch!

packages/components/src/app/word-query/word-query-adapter.js
kirlat: > I think that it is pretty the same - define a method as async and return a new Promise, I believe that async is redundant.

Yes, it is not required technically, but I've used it as a visual indication that this function executes asynchronously (i.e. returns a promise). Declaring the function `async` also changes the exception handling: any exception thrown and not caught is transformed into a rejection. However this feature is not used in the `getFromObservableQuery()`  as we have a `try/catch` block around all the code within this function.

It all is very opinionated. Here is a discussion about it on StackOverflow: https://stackoverflow.com/questions/44106907/when-to-mark-function-as-async I think we should make a decision about whether to use or not to use `async` in cases like that and stick to it.

@balmas, @irina060981, what do you think should be our approach here?

kirlat: > by the way I think that having the same names for a local variable and the whole method around of this variable is rather confusing and makes the code worse readable, also you don't even return the local variable as the result of this method

Agree with that, will rename the local variable.

kirlat: `observableQuery` is a wrapper around the Apollo Watcher that adds the following functionality:
* It introduces that maximum number of iterations that is used to terminate the long-running queries
* It checks our own custom `loading` status to verify if the query is still running
* It splits query response into three separate parts: homonyms, state, and errors thus hiding the internal structure of the GraphQL response from the client.

packages/components/src/data-model/word-query/word-query.js
balmas: yes

packages/components/src/app/word-query/word-query-adapter.js
kirlat: > and then here getFromObservableQuery - we redefine dataCallback with the different argumnets list?
Was this data.word.homonyms, data.word.state, data.word.errors and becomes this (data, loading) ?

Oh, there is an error with the arguments inside the `getFromObservableQuery()`. I've missed it because this method is not used in the current implementation. I will update it to fix the issue. Thanks for catching this!

balmas: > It all is very opinionated. Here is a discussion about it on StackOverflow: https://stackoverflow.com/questions/44106907/when-to-mark-function-as-async I think we should make a decision about whether to use or not to use `async` in cases like that and stick to it.
> 
> @balmas, @irina060981, what do you think should be our approach here?

I'm having deja vu about this one. I think we've debated about it before. I prefer to see the async keyword.

kirlat:
from https://github.com/alpheios-project/alpheios-core/pull/538#discussion_r496672962
> I think it would help to know when we expect this method to be used. In what scenarios would we want the default callback?

@balmas This method is not used in the current implementation. Should I remove it according to the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle?

balmas:
> from [#538 (comment)](https://github.com/alpheios-project/alpheios-core/pull/538#discussion_r496672962)
> 
> > I think it would help to know when we expect this method to be used. In what scenarios would we want the default callback?
> 
> @balmas This method is not used in the current implementation. Should I remove it according to the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle?

I guess it depends upon whether you expect us to need it. If you don't have a clear idea of when we might use it then I would vote for removing it.

packages/wordlist-next/README.md
```diff
@@ -0,0 +1,13 @@
+# User word-in-context lists
```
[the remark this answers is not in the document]
kirlat: It's the work-in-progress for the upcoming wordlist controller refactoring, but it is out of scope for the current PR. Please do not review it. I will eliminate this code from the reviewed branch.

packages/data-models/src/language.js
```diff
@@ -0,0 +1,70 @@
+// TODO: This creates a circular dependency and probably prevent Jest tests to succeed
+import { Lang } from './constants.js'
```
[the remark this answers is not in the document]
kirlat: That's because we import `Lang` into `language.js` from the `constants.js`. `constants.js`, in turn, imports `Language` from `language.js`. I think it is probably a bad design and I need to rethink it.

packages/data-models/src/constants.js
```diff
@@ -12,6 +14,20 @@ export const LANG_GEEZ = Symbol('ge\'ez')
...
+export const Lang = {
+  LATIN: new Language('lat'),
+  GREEK: new Language('grc'),
+  ARABIC: new Language('ara'),
+  PERSIAN: new Language('per'),
+  GEEZ: new Language('gez'),
+  CHINESE: new Language('zho'),
+  SYRIAC: new Language('syr')
+}
+
```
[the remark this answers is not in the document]
kirlat: I'm not sure myself. I was not able to find any information about approach like this beings used, but I also did not find any info about whether using it is bad. But it's probably better to rethink the design because of the circular reference problem described above.

packages/components/src/app/word-query/word-query-adapter.js
kirlat: > I'm having deja vu about this one. I think we've debated about it before.

I think we can start a "Best practices" document to put cases like that there so we won't forget. This could be the first case for it 🙂.

irina060981: I agree, that it could be used with directly defined Promise in return and in the case we return non-Promise, that would be converted to Promise by async keyword.
Thank you for the explanations.

packages/components/src/data-model/word-query/word-query.js
irina060981: I would think about the following structure (all titles are only as an example, and of course it is only my thoughts :) ):

- create a separate class jobOperation, that would have only static methods with all available jobOperations - `_getTuftsMorphology`, `_getTreebankData` and so on; and place here requirements of adding the current step - so we could see easily them. May be I would even create a separate folder with such classes , each file - is the one OperationStep, that has clear input, output, workflow and requirements in stadartized manner.

- create a separate class for ExecutionPath - and place there the hierarchy logic of combining steps, I would remove this complex if queue, that is not really readable and would be difficult to support; Here it would be good to place requirements to the order of steps, to the defining sequence or parallel they will be;

- and there will be only general methods in WordQuery class;

This way I would be able to separate layers of data and made them more independent, easier supported and easier expandable.

And this way I would have the following "bricks" for queries - JobOperation -> ExecutionPath -> Query

And also I didn't finally understand - would you have only one query - wordQuery that would retrieve all data, or would create several queries with different ExecutionsPaths inside?

If I would decide to have different Queries, for example GeneralLookupQuery, ShortDefsWordlistQuery, FullDataForWordItemQuery and so on, then I would create the following hierarchy:

 JobOperation -> ExecutionPath -> <specificQuery> -> Query
where Query would have general methods to work with AppoloClient, and <specificQuery> would have available executionPaths, requirements for input and output format for this query

@kirlat and @balmas , what do you think? Would such thoughts about code/data organization helpful?

kirlat: > I would think about the following structure

I like this suggestion, I was thinking about something along those lines. It also fits the DDD (Domain-Driven Design) very well. I will update the code accordingly.

kirlat: > And also I didn't finally understand - would you have only one query - wordQuery that would retrieve all data, or would create several queries with different ExecutionsPaths inside?

It is supposed to be one query, but with the query parameters that specify what data and from what sources should be retrieved.

Right now it uses the following parameters.
  `get` parameters specify what information should be returned:
    `getLexemes` - specifies that the lexemes need to be retrieved:
    `getShortDefs` - indicates that short definitions are needed;
  `use` parameters specify what sources should be used:
    `useMorphService` - specifies that a "main" morphology service should be used (such as Tufts);
    `useTreebankData` - specifies that data from treebank should be use;
    `useWordAsLexeme` - indicates that we should construct a lexeme out of a word;

Using different combination of parameters we can configure the query to return exactly what we need.

Does it makes sense? Are there any cases when this approach won't work?

packages/components/src/data-model/word-query/word-query-resolver.js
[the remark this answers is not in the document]
kirlat: I'll add an explanatory comment to the code

packages/components/src/app/word-query/word-query-adapter.js
kirlat: I will start to write the Best Practices document.

kirlat:
> As I understand it, the resolver of a query should be behind the graphql API. But it looks to me like we use this resolver as the read function for the InMemoryCache. And it's not clear to me where the query execution actually happens
> 
> We specify an endpoint of "/graphql" when we create the ApolloClient but that doesn't correspond to any real endpoint. So it looks like it's handing it off to the cache to do the actual resolution of the query.

That is correct, the "/graphql" endpoint is just a convention and is not really used. The query is fully local because all its fields (i.e. a slinge `word` field for now) has an `@client` (that's an indication that a resolution for this field should happen locally):
```
query WordQuery($langCode: Language, $userID: String) {
      word(langCode: $langCode, userID: $userID) @client
    }
```
So this query is resolved from the Apollo's in-memory cache.

> I would expect the cache to have a reference to an adapter method to convert the JSON of the results of the query resolution (or execution) into the HomonymGroup object, but instead it seems to responsible for actually resolving the query, which seems a little backwards.

This is how I've seen it when was designing an architecture. `WordQueryAdapter` is the last object on the client side, right in front of a separation line between the client and the server. It is responsible for communicating with the "server". It's purpose is to hide implementation details of the query from the client. `WordQueryAdapter` holds the Apollo in-memory cache that belongs to the client side: https://www.apollographql.com/docs/react/local-state/local-state-management/.

However, if we want to consider the possibility of the word query to be remote, the resolution logic should belong to the server side part of business logic. This is the `WordQueryResolver` is. It is responsible for resolving the incoming GraphQL query and it uses a `WordQuery` class for that. `WordQueryAdapter` imports the `WordQueryResolver` (because there is no other way to implement a local field resolution). I see this connection as a bridge between "local" and the "remote" code.

Does this make sense? What do you think?

packages/components/src/data-model/word-query/word-query.js
balmas: I think @irina060981 's suggestion for code organization is a good one.

Utlimately the challenge we are facing with this code is this:
1) we want to hide the complexity of how the different data sources are combined to make up the eventual word data that the user sees
2) at the same time, we want to make it possible for the user to have a say in what sources are used and how they are combined
3) and we want to be able to add to the available data sources as new resources become available

So we need to make the operations for retrieving the data as granular as possible, the algorithm for merging and disambiguating the data as flexible and self-describing as possible and the code for adding a new data source as simple as possible. It's not an easy challenge and I think it will take us several passes through to get it closer to where it needs to be.  

I believe that this is precisely the problem that the GraphQL architecture is designed to enable us to solve, but it will take some experience with it to figure out how to use it to full advantage.

balmas: regarding the question of one or multiple queries, I believe the wordquery should be one query but we may have others for other use cases.  For now, I think retrieval of full definitions and other related resources do not warrant a graphql query. I think eventually the user wordlist will but not until we have other sorts of word-related user data.

balmas:
> > As I understand it, the resolver of a query should be behind the graphql API. But it looks to me like we use this resolver as the read function for the InMemoryCache. And it's not clear to me where the query execution actually happens
> > We specify an endpoint of "/graphql" when we create the ApolloClient but that doesn't correspond to any real endpoint. So it looks like it's handing it off to the cache to do the actual resolution of the query.
> 
> That is correct, the "/graphql" endpoint is just a convention and is not really used. The query is fully local because all its fields (i.e. a slinge `word` field for now) has an `@client` (that's an indication that a resolution for this field should happen locally):
> 
> ```
> query WordQuery($langCode: Language, $userID: String) {
>       word(langCode: $langCode, userID: $userID) @client
>     }
> ```
> 
> So this query is resolved from the Apollo's in-memory cache.
> 
> > I would expect the cache to have a reference to an adapter method to convert the JSON of the results of the query resolution (or execution) into the HomonymGroup object, but instead it seems to responsible for actually resolving the query, which seems a little backwards.
> 
> This is how I've seen it when was designing an architecture. `WordQueryAdapter` is the last object on the client side, right in front of a separation line between the client and the server. It is responsible for communicating with the "server". It's purpose is to hide implementation details of the query from the client. `WordQueryAdapter` holds the Apollo in-memory cache that belongs to the client side: https://www.apollographql.com/docs/react/local-state/local-state-management/.
> 
> However, if we want to consider the possibility of the word query to be remote, the resolution logic should belong to the server side part of business logic. This is the `WordQueryResolver` is. It is responsible for resolving the incoming GraphQL query and it uses a `WordQuery` class for that. `WordQueryAdapter` imports the `WordQueryResolver` (because there is no other way to implement a local field resolution). I see this connection as a bridge between "local" and the "remote" code.
> 
> Does this make sense? What do you think?

Yes this makes much more sense to me now. Thank you for the link to the relevant Apollo Client docs.  I think your approach is a good start.  So is this correct:

*  the **WordQuery** implements the local Data fetching API.  It contains the execution logic for the query, and provides as such the "server" side of the GraphQ implementation

*  the **WordQueryResolver** is providing domain business logic for both the "client" side of the query. It decides how to execute the query, how to watch the results and store them in memory.

* the **WordQueryAdapter** is on the app side, it formulates the GraphQL query, decides where to send it, and delegates to Apollo Client to execute it.

In this architecture, we are also still using the Client Adapters to execute domain business logic to turn individual service results into domain objects.  I think this double usage of the term "Adapter" is added some confusion for me here.   I don't think the WordQueryAdapter is actually performing an adapter role. Would "WordQueryController" or "WordQueryContainer" be more accurate?

kirlat:
> Would "WordQueryController" or "WordQueryContainer" be more accurate?

I think you're right that `Adapter` is not the appropriate term to describe the `WordQueryAdapter` responsibilities. I think we can rename it to the `WordQueryController`, if you don't mind. It will fit well into the "Controller" role within the DDD terminology.

kirlat:
I will close this PR in favor of a new one, in order to keep discussion more manageable

Issue #534, opened 2020-08-21
fix and reinstate vue tooltip component test


balmas:
there is something about the Vue component tooltip test that causes the test suite to hang when it is run in the CI environments (it hangs on both Travis and GitHub Actions).

I don't think it's related to a change in the Alpheios code, because the same builds that succeeded ~> 1 week ago are failing when rerun now.  It's not a difference in a top-level dependency, because the dependencies in the package-lock.json are no different on a build that succeeded and one that failed. It could be a low-level dependency  in node_modules, but I have no idea which one.

I suspect that if we cleanup the way the vuex store is being handled in this test it should resolve it -- it looks like it wasn't updated to use the BaseTestHelp methods (I started down that path but each test case creates its own store so it needs a bit more work).

For now I have disabled the test (by making it into a backup fie).  @kirlat when you have time can you take a look at this? It's not urgent but we should fix it before our next release.

kirlat:
This should be fixed by 856d9dfd2e262882ae0383b1419ca921d2955c34. I'm still not sure why was it happening. It was probably something related to the way Vue Test Utls work internally. Maybe it takes time for each change to apply and we need to wait for the previous one to finish before applying the next one. But anyway, the fix in 856d9dfd2e262882ae0383b1419ca921d2955c34 seems to be enough to solve this.

PR #539, opened 2020-10-08
A second round of refactoring changes

kirlat:
This is an implementation of all the valuable suggestions that I've received during the first PR, and some other improvements as well. It's still not fully production ready and many tests are missing; I'm working on that.

The purpose of this PR is to offer for discussion and finalize a new architecture. Once it's set, I can add reliability improvements whenever necessary to be sure it handles all edge cases well.

The most significant changes are:
* Job steps are split into their own classes, as suggested by @irina060981. I've decided to represent them as value objects with the `retrieve()` non-static method. That decision was partially based on the [StackOverflow discussion](https://stackoverflow.com/questions/20310524/domain-driven-design-depends-on-static-methods/20311483). Adding a new source of data would require to create a new class for it and add that class the execution path. It should be pretty easy.
* First I've decided to move execution path out of the `WordQuery` but then I've realized that nothing will be left within the `WordQuery` class after that. So I decided to leave the execution path where it was, within the `WordQuery`. The `WordQuery` now, in fact, is the execution path because there is not much else there. Do you think that this is OK?
* `WordQueryResult` has been renamed to the `WordQueryResponse` to better reflect its role.
* I found a way to simplify "live" tracking of query results by using just a simple callback. This allowed to get rid of the proxied version of the `WordQueryResult` and simplified code a lot. It also provided more flexibility as we can chose the moment to call the callback, and we can do that at moment where a `WordQueryResponse` would guaranteed to be in a consistent state.
* In order to standardize communication between objects further I've decided to add a `LexicalDataResponse` class that represent a result of a singe job. This makes communication more predictable and consistent. Results of several jobs are stored in a map with the key being the job name. This map is passed to each job data object so they could use data from previous job steps if necessary.
* I've added a more standardized error handling and created a mapper between the client adapters and the GraphQL error formats. An enum class was created for error codes. However, the way errors are handled in GraphQL, error codes, messages, etc. is still up to discussion.
* `WordQueryAdapter` has been renamed to the `WordQueryController` to better reflect its purpose, upon a @balmas suggestion from a previous PR.

Please let me know what do you think. Thanks!


packages/components/src/data-model/word-query/error/word-query-error-codes.js
```diff
@@ -0,0 +1,8 @@
+/** @enum {string} */
+const WordQueryErrorCodes = {
```
kirlat, line 2: I might decide to use codes more detailed than that.

packages/components/src/data-model/word-query/error/word-query-error.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+export default class WordQueryError {
+  /**
+   * Creates an instance of WordQueryError.
+   *
+   * @param {string} message - An error message.
+   * @param {WordQueryErrorCodes} code - An error code.
+   * @param {string[]} path - A path where an error occurred.
+   */
+  constructor (message, code, { path = [] } = {}) {
```
kirlat, line 15: The error code is used because it's part of an Apollo convention for errors returned by the GraphQL query.

packages/components/src/data-model/word-query/lexical-data/result/lexical-data-result.js
```diff
@@ -0,0 +1,16 @@
...
+      throw new Error('LexicalDataResult cannot be created without a data type')
+    }
+
+    this.dataType = dataType
+    this.state = {
+      loading: false,
+      available: false,
+      failed: false
+    }
+    this.data = undefined
```
kirlat, line 13: `data` would usually contain the `HomonymGroup` but there might be something else in the future. I've decided to call it `data`, not `homonymGroup`, to keep things flexible.

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,79 @@
...
+      word: variables.word,
+      language: variables.language,
+      contextForward: variables.contextForward,
+      contextBackward: variables.contextBackward
+    })
+    // The words map will store references to the results stored in the Apollo cache
+    if (!storage.words) { storage.words = new Map() }
+    if (!storage.words.has(key)) {
+      // eslint-disable-next-line prefer-const
+      let response = new WordQueryResponse({ updateCallback: this.updateCallback.bind(this, storage, key) })
```
kirlat, line 38: Register a callback to be called when response data changes.

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,79 @@
...
+        lexicons,
+        shortLexicons,
+        wordQueryResponse: response
+      })
+      // Start the new query to obtain lexical data. Don't wait for it to resolve:
+      // it will dynamically update the result object until the query if complete fully.
+      wordQuery.start()
+
+      // This makes an Apollo reactive variable
+      const word = makeVar(response.toJsonObject())
```
kirlat, line 62: Response returned to the user is a JSON representation of a `WordQueryResponse`

packages/components/src/data-model/word-query/word-query-response.js
```diff
@@ -0,0 +1,45 @@
...
+      homonyms: this.homonymGroup ? this.homonymGroup.homonyms.map(h => h.convertToJSONObject(true)) : [],
+      state: JSON.parse(JSON.stringify(this.state))
+    }
+    if (this.errors.length > 0) {
+      result.errors = this.errors.map((err) => err.toJsonObject())
+    }
+    return result
+  }
+
+  notify () {
```
kirlat, line 40: This will call the notification callback. It will make the updated response object available to the query resolver.

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,176 @@
...
+    useTreebankData,
+    useWordAsLexeme,
+    getShortDefs,
+    treebankProvider,
+    treebankSentenceId,
+    treebankWordIds,
+    lexicons,
+    shortLexicons,
+    wordQueryResponse
+  }) {
```
kirlat, line 27: Don't like the long list of parameters but has no idea how to avoid using it.

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,176 @@
...
+    get parameters specify what information should be returned:
+        getLexemes - specifies that the lexemes need to be retrieved:
+        getShortDefs - indicates that short definitions are needed;
+    use parameters specify what sources should be used:
+        useMorphService - specifies that a "main" morphology service should be used (such as Tufts);
+        useTreebankData - specifies that data from treebank should be use;
+        useWordAsLexeme - indicates that we should construct a lexeme out of a word;
+     Using different combination of parameters the client can configure the query to return exactly what is required.
+     */
+    this._queryData = {
```
kirlat, line 39: Decided to put it into an object in order not to pollute the object props.

packages/components/src/data-model/word-query/word-query.js
```diff
@@ -0,0 +1,176 @@
...
+            if (jobResult.errors.length > 0) {
+              // Copy errors from all jobs into the query response
+              this._wordQueryResponse.errors.push(...jobResult.errors)
+            }
+            if (jobResult.dataType === DisambiguatedData.dataType) {
+              this._wordQueryResponse.state.lexemes = jobResult.state
+              this._wordQueryResponse.homonymGroup = jobResult.data
+              if (jobResult.state.failed) {
+                // Skip other steps if disambiguation failed
+                this.cancel()
```
kirlat, line 145: This will make the executor function to skip all consequent batches.

packages/data-models/src/language.js
```diff
@@ -0,0 +1,115 @@
...
+     An ISO 639-3 _code.
+     *
+     @private
+     @type {string}
+     */
+    this._code = code
+  }
+
+  static get LATIN () {
+    return new Language(Lang.LATIN)
```
kirlat, line 39: A static getter that will create an instance of a `Language` value object for a specified language.

packages/components/src/data-model/word-query/error/error-mapper.js
```diff
@@ -0,0 +1,22 @@
+import WordQueryError from '@comp/data-model/word-query/error/word-query-error.js'
+
+/**
+ * A class that maps errors from one format to the other.
+ */
+export default class ErrorMapper {
+  /**
+   * Converts an error returned by client adapters to a format compatible with GraphQL schema.
+   *
+   * @param {object} clientAdaptersError - An error returned by client adapters.
```
balmas, line 10: minor question: would this only ever be used for clientAdapter errors?  If we can envision other uses for it maybe the parameter should be more generally named.

packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -0,0 +1,79 @@
...
+    if (!storage.words) { storage.words = new Map() }
+    if (!storage.words.has(key)) {
+      // eslint-disable-next-line prefer-const
+      let response = new WordQueryResponse({ updateCallback: this.updateCallback.bind(this, storage, key) })
+      const lexicons = this._getLexiconsFn(variables.language, variables.location)
+      const shortLexicons = this._getShortLexiconsFn(variables.language, variables.location)
+      const wordQuery = new WordQuery({
+        word: variables.word,
+        language: Language.fromJsonObject({ code: variables.language }),
+        clientId: 'graphql-client',
```
balmas, line 44: I think the client Id should be more specific -- can we construct it from the application build info?

packages/components/src/data-model/word-query/word-query.js
balmas: perhaps we should subdivide them into logical groups?

useMorphService, useTreebankData useWordAsLexeme are all only relevant for getLexemes so they could be properties on the getLexemes option

treebankProvider, treebankSentenceId, treebankWord ids could be subproperties on useTreebankData

lexicons, shortLexicons could be subproperties on getShortDefs

packages/components/tests/vue/components/tooltip.test.js
```diff
@@ -50,16 +49,16 @@ describe('tooltip.test.js', () => {
...
       propsData: {
         tooltipText: ''
       }
     })
     expect(cmp.isVueInstance()).toBeTruthy()
   })
 
-  it('2 Tooltip - renders a vue instance (min requirements)', () => {
-    let storeLocal = new Vuex.Store({
+  it('2 Tooltip - renders a vue instance (min requirements)', async () => {
```
balmas, line 60: are the changes to this the fix for #534 ?

packages/components/src/data-model/word-query/lexical-data/data-objects/treebank-data.js
```diff
@@ -0,0 +1,64 @@
...
+  constructor ({ word, language, clientId, treebankProvider, treebankSentenceId, treebankWordIds } = {}) {
+    this._word = word
+    this._language = language
+    this._clientId = clientId
+    this._treebankProvider = treebankProvider
+    this._treebankSentenceId = treebankSentenceId
+    this._treebankWordIds = treebankWordIds
+  }
+
+  async retrieve () {
```
balmas, line 17: since the retrieve call is always passed a lexical data object, maybe it should be here even though it's not used (and will be empty)?

packages/components/src/data-model/word-query/lexical-data/data-objects/tufts-morphology-data.js
```diff
@@ -0,0 +1,60 @@
...
+
+export default class TuftsMorphologyData {
+  constructor ({ word, language, clientId, clearShortDefs } = {}) {
+    this._word = word
+    this._language = language
+    this._clientId = clientId
+    this._clearShortDefs = clearShortDefs
+  }
+
+  async retrieve () {
```
balmas, line 15: since the retrieve call is always passed a lexical data object, maybe it should be here even though it's not used (and will be empty)?

packages/components/src/data-model/word-query/lexical-data/data-objects/word-as-lexeme-data.js
```diff
@@ -0,0 +1,24 @@
+import LexicalDataResult from '@comp/data-model/word-query/lexical-data/result/lexical-data-result.js'
+import { Lexeme, Lemma, Homonym, HomonymGroup } from 'alpheios-data-models'
+
+export default class WordAsLexemeData {
+  constructor ({ word, language } = {}) {
+    this._word = word
+    this._language = language
+  }
+
+  async retrieve () {
```
balmas, line 10: since the retrieve call is always passed a lexical data object, maybe it should be here even though it's not used (and will be empty)?

packages/components/src/data-model/word-query/error/error-mapper.js
kirlat: If we would take other error types as an argument, the method would require a different parameter set as different errors may have different error information in them and we have to supply the missing information with the parameters. I was thinking about this method as of something to convert an `AdapterError` to a `WordQueryError`. Maybe the method should be called like `adapterToWordQueryError`? How do you think it would be better to handle this?

packages/components/src/data-model/word-query/word-query-resolver.js
kirlat: Yes, we definitely could do that. What's the purpose of the client ID on the server side? How detailed should it be? As detailed as possible?

balmas: it's used to track usage in the web server logs. So yes the more detailed the better.

packages/components/src/data-model/word-query/error/error-mapper.js
balmas: I think it's not that critical but if it's only to be used for mapping client adapter errors then I think it makes sense to make the class name reflect that.

packages/components/src/data-model/word-query/word-query-resolver.js
kirlat: Thanks for explaining, will make this change

kirlat:
> perhaps we should subdivide them into logical groups?
> useMorphService, useTreebankData useWordAsLexeme are all only relevant for getLexemes so they could be properties on the getLexemes option
> treebankProvider, treebankSentenceId, treebankWord ids could be subproperties on useTreebankData
> lexicons, shortLexicons could be subproperties on getShortDefs

Great suggestion! I hate long lists of parameters. I will update the code accordingly.

packages/components/tests/vue/components/tooltip.test.js
kirlat: I was not aware of #534 when I was making this change but I think it should fix it. I just discovered that `tooltip.test.js` is hanging even though all the tests succeeds and I decided to fix that.

packages/components/src/data-model/word-query/lexical-data/data-objects/treebank-data.js
kirlat: I'm not sure what are best practices in cases like this, but I think it makes sense. I will add a comment explaining why this variable is not used.

I also think it would make sense to create a common base class for all `Data` objects in order to define their common interface. What's your opinion on that?

packages/components/src/data-model/word-query/lexical-data/data-objects/tufts-morphology-data.js
kirlat: Will add that

packages/components/src/data-model/word-query/lexical-data/data-objects/word-as-lexeme-data.js
kirlat: Will add

packages/components/src/data-model/word-query/word-query.js
kirlat: Seen your response later on the page but not in this place. The second one was without an ability to reply so I cited it in a separate message at the very bottom.

I think it's a great idea and I will implement the suggested change.

irina060981 (approved):
@kirlat , I didn't find any tests for your changes for new workflow for LexicalQuery.
And also I could see that some tests are skipped.
Do you have an opportunity to create tests for GraphQL workflow?
To create some scenarios with different execution paths and see the way it goes through conditions with observable quiery? With different languges?

The work you've done is really big and now your new changes are mixed with your previous changes. I could see that you apply some suggestions and some suggestions you found not applicable. I believe it is time to merge your changes and see how it works.

kirlat:
> @kirlat , I didn't find any tests for your changes for new workflow for LexicalQuery.
> And also I could see that some tests are skipped.
> Do you have an opportunity to create tests for GraphQL workflow?
> To create some scenarios with different execution paths and see the way it goes through conditions with observable quiery? With different languges?

I'm working on adding more tests right now, I'm planing to merge once the tests will be ready.

PR #542, opened 2020-10-20
Small fixes for Warnings in build process

irina060981:
- fix Warnings for absent key property with v-for inside inflection-browser.vue
- fix Warnings for named import (from webpack 5)

@balmas and @kirlat these are really small cosmetic changes , so I would merge without approval.


irina060981:
There are also some errors in components thta are not reproduced locally, won't spend time for them for now - I know that @kirlat was working already on it

PR #543, opened 2020-10-20
Updated the client ID

kirlat:
Added an extended app information into the GraphQL client ID.


kirlat:
Will merge without a review because the changes are minor

PR #546, opened 2020-10-30
Fix tokenization adapter - availability error

irina060981:
For issues https://github.com/alpheios-project/alignment-editor-new/issues/50 and https://github.com/alpheios-project/alignment-editor-new/issues/54
- commented availability check for the tokenization service - we need to get availability langs first (skipped one unit test)
- added availability error to ClientAdapters initialization - tokenization service (would be published to Notification bar)

image


irina060981:
@balmas , I turned this check for available languages by default - but here I don't know what languages could be used.
And how tokenization servie is working? 
I believe it doesn't check the languages but fails for non cirillic characters?

irina060981:
@balmas, I would merge it without approval - there are small changes to fix the bug.
Tests are well (besides one lexicon test because of network traffic)

balmas:
> @balmas , I turned off this check for available languages.
> I don't know what languages could be used.
> And how tokenization servie is working?
> I believe it doesn't check the languages but fails for non cirillic characters?

The current explicitly supported languages can be found in https://github.com/alpheios-project/tokenizer/blob/master/tokenizer/lib/spacy/lang/mapper.py

If it's not one of those it will default to a generic multilanguage model

irina060981:
then there is no need in available languages check, I believe

balmas:
> then there is no need in available languages check, I believe

I agree.

PR #552, opened 2020-11-06
A dependency update

kirlat:
It is not a fix of a webextension issue per se, but rather a dependency update, including the one to the 5.x production version of webpack, to eliminate the impact of potential incompatibilities in dependencies.

Will submit a webextension fix separately.


balmas (approved):
ok

Issue #548, opened 2020-11-04
short definitions not found doesn't remove loading message


balmas:
Reported by @monzug against Alpheios Components 3.3.1 build i42-tokenization-service-fix.20201030499
(via https://github.com/alpheios-project/webextension/releases/download/dev.20201102626/dist.zip)

if a greek short def isn't found, the loading message doesn't get replaced with the "definition not found" message

Reproduce with these words:

Δαναὸς    
ἀφίκτωρ 
λεπτοψαμάθων

In release version we get "definition not found" in master, the loading message

kirlat:
It should be fixed by #554.

monzug:
to test  ἀφίκτωρ in https://scaife.perseus.org/reader/urn:cts:greekLit:tlg0085.tlg001.perseus-grc2:1-30/

PR #554, opened 2020-11-09
A fix for the short definitions not found message

kirlat:
Fixes the issue with the progress bar not disappearing after short definitions are not found (see #548).


balmas (approved):
thanks

packages/components/src/vue/vuex-modules/data/auth-module.js
```diff
@@ -133,7 +133,7 @@ AuthModule.api = (moduleInstance, store) => {
       }
       moduleInstance._auth.session().then((data) => {
         moduleInstance._api.updateAuthData(data)
-      }).catch((error) => { // eslint-disable-line handle-callback-err
+      }).catch((error) => { // eslint-disable-line node/handle-callback-err
```
irina060981, line 136: @kirlat and @balmas , I believe we have some problems with building here.
Because with this statement `node/handle-callback-err` my build process fails.
I was to remove `node/` to be able to build the package.
Why do you adding `node/` ? Do you hav the opposite problem with the build?

irina060981 (changes requested):

irina060981:
I have no questions to the code - besides eslint statement.
I believe we should use the same eslint settings in development process.
I couldn't understand why they could differ.

packages/components/src/vue/vuex-modules/data/auth-module.js
kirlat: The `node` prefix is what eslint requires me to have in order to build the package. I believe this is because either ESLint or its rules packages have been updated.

I'm wondering if you have an outdated version of dependencies. Can you please check that your ESLint/rules packages are matching those listed in the `package.json` of components? I.e.:
```
"eslint": "^7.12.1",
"eslint-config-standard": "^16.0.1",
"eslint-plugin-import": "^2.22.1",
"eslint-plugin-jsdoc": "^30.7.7",
"eslint-plugin-node": "^11.1.0",
"eslint-plugin-promise": "^4.2.1",
"eslint-plugin-standard": "^4.0.2",
"eslint-plugin-vue": "^7.1.0",
"eslint-scope": "^5.1.1"
```
I'm wondering if this issue occurred because of the dependencies mismatch. `npm run bootstrap` from the root of `alpheios-core` should probably fix that.

irina060981: Yes, it is quite the same - I uploaded your branch and didn't change anything.
I don't think that the problem is from libs versions.

kirlat: Hmm... Let me investigate this further.

irina060981: I uploaded your branch, reinstalled the dependencies - and build, for now i don't have problems with building.
I believe that there was some problems with updating dependencies through lerna.
I have no additional questions here!

irina060981 (approved):
Looks well

packages/components/src/vue/vuex-modules/data/auth-module.js
kirlat: Lerna updates can be tricky. Sometimes it updates everything, but sometimes it misses something out. I still did not figure out why is that happening 🙁, so in a moment of doubt I go to the package locations (`node_modules`) and check the versions in `package.json` manually.

Issue #550, opened 2020-11-05
lost feedback while waiting for first data after query start


balmas:
reported by @monzug against Alpheios Components 3.3.1 build i42-tokenization-service-fix.20201030499
(via https://github.com/alpheios-project/webextension/releases/download/dev.20201102626/dist.zip)

in the release, first we do show the pop-up with data is loading and then a second pop-up with the morphological data.
in the latest test env, the pop-up with data is loading is not longer generated. is this intentional? Note that the user is stuck with no information on what's going  between the moment the word is entered in lookup and the morphological pop-up is generated.

monzug:
this is again noticeable with τοῦ in
https://scaife.perseus.org/reader/urn:cts:greekLit:tlg0610.perseus002.perseus-grc1:1-5/.
(both lookup or doubleclick)

monzug:
if I double click on τοῦ  from user word list, the lexical data is loading pop-up is still generated.

kirlat:
Fixes has been merged to master

Issue #547, opened 2020-11-04
query timeout problem


balmas:
Reported by @monzug against Alpheios Components 3.3.1 build i42-tokenization-service-fix.20201030499
(via https://github.com/alpheios-project/webextension/releases/download/dev.20201102626/dist.zip)

>Chrome/Mac,   go to https://scaife.perseus.org/reader/urn:cts:greekLit:tlg0610.perseus002.perseus-grc1:1-5/
double click on   τοῦ   or lookup for τοῦ,  it loads forever. I had to close the pop-up. See screenshot.
Then, I did more testing including login in and out, did some testing on Latin then back on Greek, this same word works fine. Maybe the first time it was waiting to load the Greek database?

I think there might be a timeout bug here.  @kirlat can you investigate? ?Could it be that the graphql query might be timing out before the defs are loaded.

kirlat:
Will check it out

monzug:
![Screen Shot 2020-11-04 at 3 07 49 PM](https://user-images.githubusercontent.com/41396793/98256008-58239780-1f7e-11eb-81da-5f365b6cc6f0.png)

kirlat:
A fix has been merged to master

PR #555, opened 2020-11-11
Lexical query fixes

kirlat:
Fixes issues when the lexical query request status was not updated properly (see #547, #550). Also includes some other minor workflow improvements.


packages/components/src/data-model/word-query/word-query-resolver.js
```diff
@@ -52,6 +52,11 @@ export default class WordQueryResolver {
     if (!storage.words.has(key)) {
       // eslint-disable-next-line prefer-const
       let response = new WordQueryResponse({ updateCallback: this.updateCallback.bind(this, storage, key) })
+      // This makes an Apollo reactive variable
+      const word = makeVar(response.toJsonObject())
+      // Store the result object reference into the map
+      storage.words.set(key, word)
+
```
kirlat, line 59: `wordQuery.start()` on line 97 calls `notify()` which expects the `word` object to be in the storage. So it's better to place the `word` to the storage before the `wordQuery.start()` call.

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -892,18 +892,22 @@ export default class AppController {
       homonym.lexemes.filter((l) => l.isPopulated()).length < 1
     if (notFound && !this._store.state.app.queryStillActive) {
       let languageName
+      let languageCode
       if (homonym) {
+        languageCode = homonym.language
         languageName = this.api.app.getLanguageName(homonym.languageID).name
-      } else if (this._store.state.app.currentLanguageName) {
+      } else if (this._store.state.app.currentLanguageCode && this._store.state.app.currentLanguageName) {
+        languageCode = this._store.state.app.currentLanguageCode
```
kirlat, line 900: If we do not have homonym, we do not have the `homonym.language` either. In that case we should take language code from the `currentLanguageCode` as we do with the language name.

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1168,7 +1173,7 @@ export default class AppController {
               we could still offer the option to search the usage examples).
               Similarly, we might prompt the user to create their own analysis...
          */
-        if (data.homonym) { this.showLanguageInfo(data.homonym) }
+        this.showLanguageInfo(data.homonym)
```
kirlat, line 1176: The presence of this check was preventing the "change language" message to be displayed in the updated lexical query workflow. Situations when `homonym` is undefined can be handled by `shoeLanguageInfo()` without any issues. So there is no reason for this check to be here.

balmas (approved):
thanks.  see https://github.com/alpheios-project/alpheios-core/commit/0af17f1f36fcbdafefb56389ad9769307baa15a6#r44079993 which I think was a source of these bugs

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -892,18 +892,22 @@ export default class AppController {
...
       } else {
+        languageCode = Constants.STR_LANG_CODE_UNDEFINED
         languageName = this.api.l10n.getMsg('TEXT_NOTICE_LANGUAGE_UNKNOWN')
       }
       if (this._store.state.app.lexicalRequest.source === LexicalQuery.sources.PAGE) {
         // we offer change language here when the lookup was from the page because the language used for the
         // lookup is deduced from the page and might be wrong
         const message = this.api.l10n.getMsg('TEXT_NOTICE_CHANGE_LANGUAGE',
-          { targetWord: this._store.state.app.targetWord, languageName: languageName, langCode: homonym.language })
+          { targetWord: this._store.state.app.targetWord, languageName: languageName, langCode: languageCode })
```
irina060981, line 910: Also here there is not a direct note to the current PR, but to the whole method `showLanguageInfo`.
I think that we should divide all calculations here between `homonym` instance, LexicalQuery current status and APP state properties.
1.
```
const notFound = !homonym ||
      !homonym.lexemes ||
      homonym.lexemes.length < 1 ||
      homonym.lexemes.filter((l) => l.isPopulated()).length < 1
```
This part is about homonym, defines empty homonym and I believe could be used in different places in packages. It is better to place it as a check inside `Homonym` class.

2.
```
if (this._store.state.app.currentLanguageCode && this._store.state.app.currentLanguageName) {
        languageCode = this._store.state.app.currentLanguageCode
        languageName = this._store.state.app.currentLanguageName
      } else {
        languageCode = Constants.STR_LANG_CODE_UNDEFINED
        languageName = this.api.l10n.getMsg('TEXT_NOTICE_LANGUAGE_UNKNOWN')
      }
```
This part is about defining `currentLanguageCode` and `currentLanguageName` for the current state of the sesson and is not directly related to the homonym language data. I believe that these calculations should be scoped inside definition of  `currentLanguageCode` and `currentLanguageName`.

3. 
```
if (notFound && !this._store.state.app.queryStillActive) {
if (this._store.state.app.lexicalRequest.source === LexicalQuery.sources.PAGE) {
        // we offer change language here when the lookup was from the page because the language used for the
        // lookup is deduced from the page and might be wrong
        const message = this.api.l10n.getMsg('TEXT_NOTICE_CHANGE_LANGUAGE',
          { targetWord: this._store.state.app.targetWord, languageName: languageName, langCode: languageCode })
        this._store.commit('ui/setNotification', { text: message, important: true, showLanguageSwitcher: true })
      } else {
        // if we are coming from e.g. the lookup or the wordlist, offering change language
        // here creates some confusion and the language was explicit upon lookup so it is not necessary
        const message = this.api.l10n.getMsg('TEXT_NOTICE_NOT_FOUND',
          { targetWord: this._store.state.app.targetWord, languageName: languageName, langCode: homonym ? homonym.language : '' })
        this._store.commit('ui/setNotification', { text: message, important: true, showLanguageSwitcher: false })
      }
    } else if (!this._store.state.app.queryStillActive) {
      this._store.commit('ui/resetNotification')
    }
```
And only this part is about showing ui message with languageInfo.
And here we could make this code more readable:
instead of this
```
{ targetWord: this._store.state.app.targetWord, languageName: languageName, langCode: languageCode }
```
do this (rename `languageCode ` to `langCode` or change it inside message definition)
```
const targetWord = this._store.state.app.targetWord
const message = this.api.l10n.getMsg('TEXT_NOTICE_CHANGE_LANGUAGE',
          { targetWord, languageName, languageCode })
```

I think that with applying my suggestions we could 
- reduce code amount without loosing its functionality
- reduce potential duplicate code 
- make code more readable

I believe that this my inspiration of code optimize is caused by yesterdays afforts to find out how we handle with options through a lot of long paths of code in `components` and `data-models`. May be you will find my suggestions helpful.
Anyway they don't prevent me from approving the code.

irina060981 (approved):
Looks well

kirlat:
> May be we should define languageName the same way as languageID, languageCode, languageModel inside data-models and make a shotter path to it this way?

I agree that language handling became way to convoluted and needs a simplification. I think the new `Language` class should replace both `languageID` and `languageCode`. If we need a string representation of a language code, we can use the `toCode()` method of the `Language`. Maybe it even worth to create a `code()` alias to it so that it will read better (`language.code()` instead of `language.toCode()`, the latter one was created for serialization purposes and the name reflects that).

The language name, I think, is probably the part of the UI logic because it can differ depending on circumstances: the name of the same language can be displayed differently in different parts of the UI. It also might be localized. So I think the logic to determine the language name should belong to the UI layer, not to the more abstract data models. So maybe we can use the function within the app controller to return the language name, as currently, but make it to take the new `Language` object instead.

For legacy logic that still requires `languageID` we can use a method of the `LanguageModel` that will get the ID out of the `Language` class.

@balmas, @irina060981, what do you think about the plan? If it is usable, I can create an issue to track that.

kirlat:
```
 const notFound = !homonym ||
      !homonym.lexemes ||
      homonym.lexemes.length < 1 ||
      homonym.lexemes.filter((l) => l.isPopulated()).length < 1
```
> This part is about homonym, defines empty homonym and I believe could be used in different places in packages. It is better to place it as a check inside Homonym class.

Agree with this. Maybe we can call the method `isEmpty()`? @balmas, what do you think?

irina060981:
> The language name, I think, is probably the part of the UI logic because it can differ depending on circumstances: the name of the same language can be displayed differently in different parts of the UI. It also might be localized. 

I use it inside Alignment Editor and it is localized in my interface, not through data-models. And from my point of view - we could use here aliases with default name and message_name for l10n. But in most cases `lat` - is Latin and so on :)

kirlat:
```
if (this._store.state.app.currentLanguageCode && this._store.state.app.currentLanguageName) {
        languageCode = this._store.state.app.currentLanguageCode
        languageName = this._store.state.app.currentLanguageName
      } else {
        languageCode = Constants.STR_LANG_CODE_UNDEFINED
        languageName = this.api.l10n.getMsg('TEXT_NOTICE_LANGUAGE_UNKNOWN')
      }
```
> This part is about defining currentLanguageCode and currentLanguageName for the current state of the sesson and is not directly related to the homonym language data. I believe that these calculations should be scoped inside definition of currentLanguageCode and currentLanguageName.

I agree with this suggestion in general, but maybe we should postpone this because with the lexical query refactoring we would probably change the way the session is handled.

kirlat:
> do this (rename languageCode to langCode or change it inside message definition)

```
const targetWord = this._store.state.app.targetWord
const message = this.api.l10n.getMsg('TEXT_NOTICE_CHANGE_LANGUAGE',
          { targetWord, languageName, languageCode })
```

Great suggestion! I will update the code

PR #557, opened 2020-11-16
Add upload default parameters from tokenization service

irina060981:
Added getCofig method to AlpheiosTokenizationAdapter

It is a step for the https://github.com/alpheios-project/alignment-editor-new/issues/32


packages/client-adapters/src/adapters/tokenization/adapter.js
```diff
@@ -47,16 +48,83 @@ class AlpheiosTokenizationAdapter extends BaseAdapter {
...
+
+  /**
+   *
+   * @param {Object} configData Response from config fetch request
+   * @param {String} textType - tei/text
+   */
+  convertToOptions (configData, textType) {
+    const configDataPath = configData.paths[`/tokenize/${textType}`].post
+
+    const exludeParameters = ['lang', 'direction']
```
balmas, line 93: why do you exclude these?

balmas (approved):
it looks ok

packages/client-adapters/src/adapters/tokenization/adapter.js
irina060981: Because it won't be defined through application settings.
It would be defined through settings inside each aligned request.

These are for defining setting in application options.

PR #563, opened 2020-11-24
Fix bug with tokenization adapter parameters

irina060981:
As some parameters could have `0` and `false` as valid options, I was to fix the comparision to not loose such values.


irina060981:
@balmas and @kirlat , this change is really tiny - I would merge it without a review.

PR #564, opened 2020-11-30
Options update for alignment editor

irina060981:
For the https://github.com/alpheios-project/alignment-editor-new/issues/32

1. updated version parse to support versions like "1.0.0"
2. added ability to upload values from external array (it is needed to upload langs array with l10n support)
3. added ability to clone options
4. added get method to format options 
5. added tests for new features


packages/data-models/dist/alpheios-data-models.node.js
```diff
@@ -10827,7 +10827,7 @@ class Options {
     try {
       parsed = {
         domain: domain,
-        version: parseInt(version),
+        version: version,
```
irina060981, line 10830: I need it to support versions of the following format: "1.0.0".

packages/data-models/src/options/options.js
```diff
@@ -24,7 +24,7 @@ export default class Options {
 
     this.defaults = defaults
     this.domain = defaults.domain
-    this.version = defaults.version
+    this.version = defaults.version.toString()
```
balmas, line 27: is there any place we reference version as an int (e.g. to compare 2 versions?)

balmas (approved):
it looks ok, just a couple of questions

packages/data-models/src/options/option-item.js
```diff
@@ -110,4 +110,13 @@ export default class OptionItem {
...
     )
   }
+
+  /**
+   *
+   * @param {Array[Object]} valuesArr - Array[option's values]
+   */
+  uploadValuesFromArray (valuesArr) {
+    this.values = [...valuesArr]
+    this.defaultValue = this.values[0].value
```
kirlat, line 120: If the first value will always be the default one, maybe it's better to create a getter for it and avoid an assignment here?

packages/data-models/src/options/options.js
```diff
@@ -137,4 +137,51 @@ export default class Options {
...
     return parsed
   }
+
+  /**
+   * Converts optionItems to the object: { name of the option: currentValue }
+   *
+   * @returns {object}
+   */
+  get formatLabelValueList () {
+    const result = {}
```
kirlat, line 147: I think it should be `let` according to what we agreed upon previously

kirlat (approved):
Looks good, I have just a couple of minor suggestions

packages/data-models/tests/options/option-item.test.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-env jest */
 import OptionItem from '@/options/option-item'
+import StorageAdapter from '@/storages/storage-adapter.js'
```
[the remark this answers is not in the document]
irina060981: It is not very neccesary, I used StorageAdapter because it was widely used inside `options.test.js`.
I would fix according to other tests inside `option-item.test.js` and remove StorageAdapter import.

packages/data-models/src/options/options.js
irina060981: It is defined as Number inside: 
`feature-options-defaults.json`
`language-options-defaults.json`
`site-options.json`
`ui-options-defaults.json`

in components packages.

I wanted simply to fix and convert it to String inside options's jsons. But after that I decided to add ability to support both variants.

packages/data-models/src/options/option-item.js
irina060981: No, I don't agree. 
In an odinary case we upload defaultValue from the local json or remote API json.
And it could be defined not from value's list.
That's why I decided to save the current option definition workflow as it used now in both applications - `alpheios-core` and `alpheios-alignment-editor`

packages/data-models/src/options/options.js
irina060981: Oh, I forgot to add es-lint exception, thank you for the catch - will fix

packages/data-models/src/options/option-item.js
irina060981: @kirlat , I think , if you want to suggest some refactoring of Options definition I think it would be better to create a new issue  for that. And make such a refactoring in a separate task.

irina060981:
all tests are passed successfully - will merge it without waiting for travis

packages/data-models/src/options/options.js
balmas: ok

packages/data-models/tests/options/option-item.test.js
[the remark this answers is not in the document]
balmas: thanks

Issue #562, opened 2020-11-19
morph adapter errors shouldn't fail the query if there are also valid results


balmas:
reported my @monzug . Reproducible with a lookup of the Latin word 'caro'

For this word, the test on `  if (adapterMorphRes.errors.length === 0) { ` at https://github.com/alpheios-project/alpheios-core/blob/master/packages/components/src/data-model/word-query/lexical-data/data-objects/tufts-morphology-data.js#L50 returns false because there are some errors in the morph response but there are also valid lexemes.

We should not entirely fail the query in this case, but proceed with the valid data results.

monzug:
other words that are failing: nero, qui, quibuscum, quis and many more

kirlat:
@balmas, @irina060981: To handle situations like that gracefully we should probably start to use warnings along with errors in our GraphQL API. I think those could have the following meanings:
* **Warning**. Something went wrong, but the error is not critical and we can continue the normal execution despite what happened. We should probably take notice, maybe report the issue in the logs, and maybe tweak an execution path a little bit to accommodate for the issue.
* **Error**. The problem is so serious that we cannot continue the normal execution under any circumstances and need to bail out. No morphological results will be shown to the user and, most likely, a user will be presented with the error message in the UI that will explain the situation somehow.

What do you think about an approach like the above? If it seems reasonable, I can add it into the GraphQL interface and, possibly, to other places.

balmas:
This makes sense to me.

irina060981:
I like this suggestion too

monzug:
Kirlat, Question: in the case of "caro" a warning will be issued. for the user, the warning is totally transparent, correct?

kirlat:
> Kirlat, Question: in the case of "caro" a warning will be issued. for the user, the warning is totally transparent, correct?

Yes, the warning will be used internally within the system but will not be shown to the user. Users will see no special messages in this case. The UX of the user would be exactly the same as in cases where things will go smoothly and there will be no warnings at all.

The user will be notified only of errors that prevent morphological data from being displayed. Such notifications would explain to the user why no data for the user's request was retrieved.

monzug:
Thanks!!!

kirlat:
[...]

PR #565, opened 2020-12-02
Added support for the lexical query failure

kirlat:
This build fixes the #562 and adds some related improvements to the GraphQL API.

Here is the short summary of changes:
1. The `word` GraphQL query response has been updated according to the results of the discussion in #562. If the query failed completely, then the `homonyms` field in the response will be `null` and there will be errors in the `error` field. If the query failed only partially, then the `homonyms` field will contain homonyms data and the `errors` field will contain warnings.
2. Warnings were added to both client adapters and the GraphQL backend.
3. I've started to add error codes to more error and warning objects. I think error codes are important if we want to identify what type of error it is (i.e. `if (errCode ===A) {}`) and act accordingly. The good thing about error codes is that they can be transferred from object to object and even across the network.


packages/client-adapters/src/errors/adapter-error.js
```diff
@@ -13,14 +13,6 @@ class AdapterError extends Error {
...
     }
   }
-
-  update (config) {
-    this.adapter = `${config.category}.${config.adapterName}`
-    this.methodName = config.method
-
-    this.message = `${this.message} (${this.adapter}.${this.methodName})`
-    return this
-  }
```
kirlat, line 23: I did not find it to be used anywhere so I've decided to remove the unused code. Please let me know if this code is needed and I will return it back.

packages/client-adapters/src/errors/remote-error.js
```diff
@@ -5,12 +5,4 @@ export default class RemoteError extends Error {
...
     this.errorCode = errorCode
   }
-
-  update (config) {
-    this.adapter = `${config.category}.${config.adapterName}`
-    this.methodName = config.method
-
-    this.message = `${this.errorCode}: ${this.message} (${this.adapter}.${this.methodName})`
-    return this
-  }
```
kirlat, line 15: Same is above, I think it is not used any more.

packages/client-adapters/src/errors/warning-codes.js
```diff
@@ -0,0 +1,8 @@
+/** @enum {string} */
+const WarningCodes = {
+  /*
+  `hdwd` field of the response of the morphological adapter contains a number, not a string
+   */
+  NO_LEMMA_IN_HDWD: 'NO_LEMMA_IN_HDWD'
+}
```
kirlat, line 7: I think it is good to have a centralized list of error codes. This way the client will know what kinds of errors to expect from the library (this file can be imported by the client app if necessary).

packages/components/src/app/word-query/word-query-controller.js
```diff
@@ -80,15 +80,15 @@ export default class WordQueryController {
...
+        const loading = data.word.extensions.state.loading
         if (loading === false) {
           watchableQuery.stopPolling()
         }
         if (counter >= maxIterations) {
           watchableQuery.stopPolling()
         }
         counter++
-        dataCallback(data.word.homonyms, data.word.state, data.word.errors)
+        dataCallback(data.word)
```
kirlat, line 91: I think it's better to pass the whole response the the calling function rather than parse it here: it might know better how to deal with the data returned.

packages/components/src/data-model/constants/error-codes.js
```diff
@@ -1,9 +1,12 @@
...
-const WordQueryErrorCodes = {
+const ErrorCodes = {
   TUFTS_ERROR: 'TUFTS_ERROR',
   TREEBANK_ERROR: 'TREEBANK_ERROR',
   LEXICONS_ERROR: 'LEXICONS_ERROR',
   DISAMBIGUATION_ERROR: 'DISAMBIGUATION_ERROR',
   SHORT_DEFS_ERROR: 'SHORT_DEFS_ERROR'
 }
-export default WordQueryErrorCodes
+export default ErrorCodes
```
kirlat, line 12: I think this list of error codes can belong not to WordQuery only but can be shared across other queries. Hence the renaming.

packages/components/src/data-model/constants/error-origins.js
```diff
@@ -0,0 +1,9 @@
+/*
+All possible origins of errors and warnings that can be returned from the GraphQL API.
+ */
+/** @enum {string} */
+const ErrorOrigins = {
+  CLIENT_ADAPTERS: 'ClientAdapters',
+  DISAMBIGUATED_DATA_OBJECT: 'DisambiguatedDataObject'
+}
+export default ErrorOrigins
```
kirlat, line 9: These are the modules where errors and warnings might be produced.

packages/components/src/data-model/constants/error-severity-types.js
```diff
@@ -0,0 +1,9 @@
+/*
+All possible severity values that can be returned from the GraphQL API.
+ */
+/** @enum {string} */
+const ErrorSeverityTypes = {
+  ERROR: 'Error',
+  WARNING: 'Warning'
+}
+export default ErrorSeverityTypes
```
kirlat, line 9: This is a way to distinguish between an error and a warning within the GraphQL `error` object (as GraphQL specs do not support a `warning` type).

packages/components/src/data-model/word-query/error/error-mapper.js
```diff
@@ -1,23 +1,36 @@
...
+        clAdapter: clientAdaptersErrorOrWarning.adapter,
+        clAdapterMethod: clientAdaptersErrorOrWarning.methodName
+      }
+    ]
+    if (clientAdaptersErrorOrWarning.constructor.name === 'AdapterError') {
+      return new WordQueryError(...params)
+    } else if (clientAdaptersErrorOrWarning.constructor.name === 'AdapterWarning') {
+      return new WordQueryWarning(...params)
+    } else {
+      throw new Error(`Unsupported error type: ${clientAdaptersErrorOrWarning.constructor.name}`)
```
kirlat, line 33: I think we need this for development to catch unsupported error objects (do not have any now). The exception will force us to fix the problem.

packages/components/src/data-model/word-query/error/word-query-warning.js
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Creates an instance of WordQueryError.
+   *
+   * @param {string} message - An error message.
+   * @param {import('../../constants/error-codes.js').ErrorCodes} code - An error or a warning code.
+   * @param {object} [warnData] - Additional parameters of the warning.
+   */
+  constructor (message, code, warnData = {}) {
+    super(message, code, warnData)
+    this.extensions.severity = ErrorSeverityTypes.WARNING
```
kirlat, line 18: Same as error, but with a different severity.

packages/components/src/data-model/word-query/word-query-response.js
```diff
@@ -22,11 +22,38 @@ export default class WordQueryResponse {
...
     // eslint-disable-next-line prefer-const
     let result = {
-      homonyms: this.homonymGroup ? this.homonymGroup.homonyms.map(h => h.convertToJSONObject(true)) : [],
-      state: JSON.parse(JSON.stringify(this.state))
+      data: {
+        homonyms: this.homonymGroup ? this.homonymGroup.homonyms.map(h => h.convertToJSONObject(true)) : []
+      },
+      extensions: {
+        state: JSON.parse(JSON.stringify(this.state))
+      }
```
kirlat, line 56: The response format has been updated to match the latest GraphQL specs.

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -457,15 +460,15 @@ export default class Lexis extends Module {
...
+        wqData = await this.getWordQueryData({
           variables,
           source
         })
       } catch (err) {
         Logger.getInstance().error('Observable word query error', err)
       }
 
-      if (homonym) {
+      if (wqData && wqData.homonym) {
```
kirlat, line 471: If there is something returned in the `homonym` field, request succeeded fully or partially.

balmas (changes requested):
This mostly looks good to me, I just had some small questions and change requests. 

I'm also wondering if it's possible to add an integration test that tests this specific bug, i.e. creating a fixture in the alpheios-fixtures package, and testing that a query on 'caro' succeeds.  I don't think we have integration tests for the new lexical query code yet, but having this sort of test for the inflection-tables has been enormously helpful (actually essential) to facilitating refactoring of that code. I think we need to start having similar tests here too. E.g. see the verified-issues.test.js in inflection-tables/tests/data

irina060981:
@kirlat and @balmas, I have a question here about wordList workflow.
We could have such a case, when the morph adapter doesn't return the result, but we need to have an empty homonym (or may be some other format) to save this word to the wordList with the current context.

For example (from embed-lib index.html), 
image

No, I am not sure, but I believe we didn't get any homonym and it is not saved to the wordlist now.
Do we decide not to track such words in wordlist with only the context?

irina060981:
The second case with Persian words.
There was done a workflow with a warning before.
When we didn't get lexical results , we show a notification for the user that he/she should wait for definitions.

For now we lost this behaviour, I couldn't found my PR with this change ad couldn't even find the notice with such a text.
Do we remove this feature?

irina060981:
Also I have a question
now we have two progress bars at the same time in popup
image

I believe the second one is for short definitions, but what does mean the first progress? According to my case in the screenshot all morph data is retrieved from the source

packages/client-adapters/src/errors/warning-codes.js
irina060981: I have a question, we have two similiar classes now
`AdapterError` and `AdapterWarning`

We use errorCode from the response for `RemoteError` and errorCode for `AdapterWarning`.
ErrorCodes for RemoteErrors you defined inside `alpheios-messaging`
ErrorCodes for Warnings you defined inside `client-adapters` packages

So this way we need to know exactly the class source of the errorCode to be able to analyze it.
Because otherwises - it is only a String and not any short code.
May be it is better to think how to unify it the way you could retrieve all the data about the error from the error source? Similiar to Response errors from web server (404, 500 and so on)?

irina060981 (changes requested):
I don't have any code notes , but have some questions about the final result.

kirlat:
> We could have such a case, when the morph adapter doesn't return the result, but we need to have an empty homonym (or may be some other format) to save this word to the wordList with the current context.

> The second case with Persian words.
> There was done a workflow with a warning before.
> When we didn't get lexical results , we show a notification for the user that he/she should wait for definitions.

Thanks for catching this! I've missed those two cases during my testing. I will review them and see what's going on. I think we would need to create integration tests for them, if possible.

> For now we lost this behaviour, I couldn't found my PR with this change ad couldn't even find the notice with such a text.
> Do we remove this feature?

I don't remember removing it specifically. @irina060981, can you recall the number of your PR? I can try to track changes from it to see what happened with the code since then.

packages/client-adapters/src/errors/warning-codes.js
kirlat: I was having similar thoughts myself. I think we should somehow harmonize errors between different parts of the application, and think about localization of error messages. That could be a topic requiring some thought and discussion. It's probably better to create an issue for that. What do you think?

packages/client-adapters/src/errors/adapter-error.js
kirlat: That's a good idea, I will do that.

packages/client-adapters/src/errors/remote-error.js
kirlat: Will mark it as deprecated.

packages/components/src/data-model/word-query/error/word-query-error.js
```diff
@@ -14,34 +21,48 @@ export default class WordQueryError {
...
+      throw new Error(this.constructor.errMsgs.NO_MESSAGE)
     }
-    if (!code) {
-      throw new Error('WordQueryError requires a code to be provided')
+    if (!errCode) {
+      throw new Error(this.constructor.errMsgs.NO_ERROR_CODE)
     }
     this.message = message
-    this.path = path
+    this.path = ['word']
```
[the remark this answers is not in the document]
kirlat: I've changed it to be more in line with the GraphQL specs. According to those, it should be a path within the GraphQL response to the elements that failed (see http://spec.graphql.org/draft/#example-90475 as an example). Since we do not have an exact element that failed (but may have that in the future) I decided to use the generic `word` element. No looking at the example more closely I realize that it would be better to use `homonyms` instead. I will do that change, if you don't mind.

Ideally we should identify the exact homonym and the lexeme that failed, but I'm not sure how easy will it be to do now.

packages/client-adapters/src/errors/warning-codes.js
irina060981: I think it is worth to be placed to a separate issue.

kirlat:
> now we have two progress bars at the same time in popup
> image
> 
> I believe the second one is for short definitions, but what does mean the first progress? According to my case in the screenshot all morph data is retrieved from the source

The upper progress bar is for the overall lexical query request. It disappears when the `LEXICAL_QUERY_COMPLETE` event is published at the `finalize()` method of the `LexicalQuery`, or in some other places if request fails.

The lower progress bar is for short definitions. It disappears when either short definitions are retrieved or when we figure out that there are no short definitions for this word.

@balmas, what would be the desired UI here? Both bars could make sense (especially if to change the wording a little bit), or we can show one or the other.

packages/components/src/data-model/word-query/error/word-query-error.js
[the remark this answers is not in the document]
balmas: >I will do that change, if you don't mind.

ok thanks

balmas:
> @kirlat and @balmas, I have a question here about wordList workflow.
> We could have such a case, when the morph adapter doesn't return the result, but we need to have an empty homonym (or may be some other format) to save this word to the wordList with the current context.
> 
> For example (from embed-lib index.html),
> image
> 
> No, I am not sure, but I believe we didn't get any homonym and it is not saved to the wordlist now.
> Do we decide not to track such words in wordlist with only the context?

this is a good point. What should happen when you lookup a word without results is that the word should be added to the wordlist with itself as a lemma -- e.g. in these lines of the production lexical query, if we have no result form the morph adapter, we create a homonym with the form itself as the lemma:
https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/queries/lexical-query.js#L181-L182

balmas:
> > now we have two progress bars at the same time in popup
> > image
> > I believe the second one is for short definitions, but what does mean the first progress? According to my case in the screenshot all morph data is retrieved from the source
> 
> The upper progress bar is for the overall lexical query request. It disappears when the `LEXICAL_QUERY_COMPLETE` event is published at the `finalize()` method of the `LexicalQuery`, or in some other places if request fails.
> 
> The lower progress bar is for short definitions. It disappears when either short definitions are retrieved or when we figure out that there are no short definitions for this word.
> 
> @balmas, what would be the desired UI here? Both bars could make sense (especially if to change the wording a little bit), or we can show one or the other.
> 
> We can, for example, use the upper bar for the lexeme retrieval and it can say something like "Lexemes are loading" and then the lower bar for short definitions ("Short definitions are loading"). First the upper bar will disappear and then the lower bar will follow suit, when short definitions' retrieval will be complete. This might give users a more detailed insight about what's going on behind the scenes.

 I would defer to @monzug on what should happen in the UI here.

packages/client-adapters/src/errors/warning-codes.js
balmas: +1

monzug:
"lexical data is loading" message is generated while loading the morphology pop-up. when morphology  pop-up is generated the message does not show up anymore. In the case of λεπτοψαμάθων  it does briefly co-exist with the loading of the short definition progress bar  in some cases (e.g. first look  up of a Greek word) then it does quickly fade away. I am ok with it.

balmas:
> "lexical data is loading" message is generated while loading the morphology pop-up. when morphology pop-up is generated the message does not show up anymore. In the case of λεπτοψαμάθων it does briefly co-exist with the loading of the short definition progress bar in some cases (e.g. first look up of a Greek word) then it does quickly fade away. I am ok with it.

Thanks @monzug . So @kirlat let's leave this for now.

kirlat:
> this is a good point. What should happen when you lookup a word without results is that the word should be added to the wordlist with itself as a lemma -- e.g. in these lines of the production lexical query, if we have no result form the morph adapter, we create a homonym with the form itself as the lemma:
> https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/queries/lexical-query.js#L181-L182

I've examined the code carefully and it seems that the WordList controller does not listen to the `MORPH_DATA_READY` event at all. It subscribes to `HOMONYM_READY` instead: https://github.com/alpheios-project/alpheios-core/blob/production/packages/wordlist/src/controllers/wordlist-controller.js#L13-L21

In the LexicalQuery a Homonym is constructed even if the morph data request fails (at the point that Bridget referenced in the quote above), and then, a few lines below, this Homonym object used in the `HOMONYM_READY` event: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/queries/lexical-query.js#L217-L247.

So it seems that in the current workflow, even if the morphological query failes, we still create a homonym from a target word nonethless nonetheless and, depending on the configuration, try to retrieve short definitions, full definitions, lemma translations, and word usage examples. Should we keep it this way? If so, for failed request the new workflow should create a Homonym, and pass it to the Lexical Query. Not sure how to be with short definitions in this case since those are retrieved behind the GraphQL facade. Should the GraphQL backend create a Homonym even if no morph data is available and try to retrieve short definitions for it?

Analyzing the code, I've discovered that `MORPH_DATA_READY` event is kind of a duplicate of `HOMONYM_READY` in our current workflow. There is only one subscriber for this event: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/controllers/app-controller.js#L1150-L1152. All it does is updates the text message. I think we can move the message update to the `onHomonymReady()` method and get rid of `MORPH_DATA_READY` completely.

@balmas, @irina060981 please let me know what do you think about the desired workflow.

kirlat:
> > "lexical data is loading" message is generated while loading the morphology pop-up. when morphology pop-up is generated the message does not show up anymore. In the case of λεπτοψαμάθων it does briefly co-exist with the loading of the short definition progress bar in some cases (e.g. first look up of a Greek word) then it does quickly fade away. I am ok with it.
> 
> Thanks @monzug . So @kirlat let's leave this for now.

Maybe we should change the text for the lower progress bar to `Short definitions are loading`? It might be more understandable what this bar is for this way. What do you think?

balmas:
> > > "lexical data is loading" message is generated while loading the morphology pop-up. when morphology pop-up is generated the message does not show up anymore. In the case of λεπτοψαμάθων it does briefly co-exist with the loading of the short definition progress bar in some cases (e.g. first look up of a Greek word) then it does quickly fade away. I am ok with it.
> > 
> > 
> > Thanks @monzug . So @kirlat let's leave this for now.
> 
> Maybe we should change the text for the lower progress bar to `Short definitions are loading`? It might be more understandable what this bar is for this way. What do you think?

Yes ! I think that's a very good idea.  Thank you!

monzug:
Absolutely, YES

kirlat:
> Yes ! I think that's a very good idea. Thank you!
> Absolutely, YES

Will update the text accordingly.

balmas:
> I've examined the code carefully and it seems that the WordList controller does not listen to the `MORPH_DATA_READY` event at all. It subscribes to `HOMONYM_READY` instead: https://github.com/alpheios-project/alpheios-core/blob/production/packages/wordlist/src/controllers/wordlist-controller.js#L13-L21
> 
> In the LexicalQuery a Homonym is constructed even if the morph data request fails (at the point that Bridget referenced in the quote above), and then, a few lines below, this Homonym object used in the `HOMONYM_READY` event: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/queries/lexical-query.js#L217-L247.
> 
> So it seems that in the current workflow, even if the morphological query failes, we still create a homonym from a target word nonethless nonetheless and, depending on the configuration, try to retrieve short definitions, full definitions, lemma translations, and word usage examples. Should we keep it this way? If so, for failed request the new workflow should create a Homonym, and pass it to the Lexical Query. Not sure how to be with short definitions in this case since those are retrieved behind the GraphQL facade. Should the GraphQL backend create a Homonym even if no morph data is available and try to retrieve short definitions for it?
> 
> Analyzing the code, I've discovered that `MORPH_DATA_READY` event is kind of a duplicate of `HOMONYM_READY` in our current workflow. There is only one subscriber for this event: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/controllers/app-controller.js#L1150-L1152. All it does is updates the text message. I think we can move the message update to the `onHomonymReady()` method and get rid of `MORPH_DATA_READY` completely.
> 
> @balmas, @irina060981 please let me know what do you think about the desired workflow.

Per discussion in Slack, the current production code is a bit of a hack, that allows us to continue to lookup a word in the additional resources (short definitions, full definitions, usage examples, etc) even if we fail to parse a lemma in the morphology service.  

I think we still want to do that, but I want to expand the data model so that we don't pretend the form is a "lemma" .  This is described further in https://github.com/alpheios-project/documentation/issues/33

kirlat:
> Analyzing the code, I've discovered that MORPH_DATA_READY event is kind of a duplicate of HOMONYM_READY in our current workflow. There is only one subscriber for this event: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/controllers/app-controller.js#L1150-L1152. All it does is updates the text message. I think we can move the message update to the onHomonymReady() method and get rid of MORPH_DATA_READY completely.

@balmas, @irina060981: do you think we can remove the `MORPH_DATA_READY` event?

irina060981:
>For now we lost this behaviour, I couldn't found my PR with this change ad couldn't even find the notice with such a text.
Do we remove this feature?
I don't remember removing it specifically. @irina060981, can you recall the number of your PR? I can try to track changes from it to see what happened with the code since then.

I couldn't find the PR it was rather long ago.
But I found that it was produced inside 
```
  onMorphDataNotFound () {
    this._store.commit('ui/setNotification', { text: this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_NOTFOUND'), important: true })
    this._store.commit('app/setQueryStillActive', true)
  }
```

>@balmas, @irina060981: do you think we can remove the MORPH_DATA_READY event?

We have a workflow for `MORPH_DATA_NOTAVAILABLE`. 
If we don't need this message:
```
  onMorphDataReady () {
    this._store.commit('ui/addMessage', this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_READY'))
  }
```
I believe we could remove it.

kirlat:
In our current workflow the `HOMONYM_READY` is published right after the `MORPH_DATA_READY` event. So if we'll move 
`this._store.commit('ui/addMessage', this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_READY'))` to the beginning of the `onHomonymReady()` handler the result should be almost the same, and we'll be able to simplify our workflow a little bit with that.

kirlat:
As discussed with @balmas, I will merge the current PR with the current changes and two small additional fixes, and will implement other changes discussed as separate PR(s). This is to prevent discussion in the current PR from becoming too big.

Issue #566, opened 2020-12-03
Refactor the resource retrieval workflow


kirlat:
Based on https://github.com/alpheios-project/alpheios-core/pull/565#issuecomment-738173271:
The overall resource retrieval workflow should become more transparent. One of the steps to achieve that would be to separate the retrieval of resources from the `LexicalQuery` (a single method currently) into separate steps. Each step should match a retrieval of a separate resource. Once steps are separated, they could be moved behind the GraphQL facade. We should also make conditions on which a resource is retrieved or not retrieved more straightforward and trackable.

kirlat:
@balmas, @irina060981: it seems that in the lexical query we create a list to record requests to get full definitions and do not run the retrieval query if one has already started: https://github.com/alpheios-project/alpheios-core/blob/production/packages/components/src/lib/queries/lexical-query.js#L25. We do not, however, remove the query from this map ever. So if the full definition for a word was retrieved before, it seems to never be re-retrieved again, even though it might not be stored and available. Do you remember why is it so? Is it to avoid running (possible) lengthy full definitions queries for the same word in parallel when user clicked the same word twice?

irina060981:
No, I don't remember this feature.
I don't like to use Maps, that's why I doubt that it is my code.

kirlat:
I think the `startedDefinitionsQueries` has something to do with caching within the client adapters: https://github.com/alpheios-project/alpheios-core/commit/c51481a60c770fc94ebdcf1ae3d57470b979edfa. It seems that requests are cached within the client adapters, and if retrieval of the definition resource has already started there, we don't want to issue another request. We'll just wait for the resource to be retrieved. Once the retrieval is complete, the client adapters will use either `FULL_DEFS_READY` or `FULL_DEFS_NOT_FOUND` to notify subscribers of the results received. That makes sense and I think we should leave it as it is for now. I'll add a comment into the `LexicalQuery` about that so that I won't have questions about it next time 🙂.

kirlat:
[...]

Issue #545, opened 2020-10-29
remote dictionary search feature


balmas:
to address the persian issues #493 and #165 we want to implement a dictionary search feature using the apis of https://dsal.uchicago.edu 

This could also allow us to search other remote dictionary resources inluding:

* Logeion
* Sanskrit dictionaries at https://github.com/cceh/c-salt_sanskrit_data

See query use case described at https://github.com/alpheios-project/documentation/issues/33#issuecomment-670571372

PR #568, opened 2020-12-08
Lexical query workflow improvements

kirlat:
This is a fix for the #562. It also tries to make architecture more flexible in order to satisfy #566 requirements. Resource retrieval for lemma translations, word usage examples, and short definitions have been separated into its own functions and moved into the Lexis. They do not have dependency on the LexicalQuery any more.

There were also some minor code improvements. Dependencies were updated to the latest versions, whenever it was possible without breaking the existing functionality.


packages/components/src/lib/controllers/app-controller.js
```diff
@@ -186,7 +186,6 @@ export default class AppController {
 
     // Subscribe to LexicalQuery events
     LexicalQuery.evt.LEXICAL_QUERY_COMPLETE.sub(appController.onLexicalQueryComplete.bind(appController))
-    LexicalQuery.evt.MORPH_DATA_READY.sub(appController.onMorphDataReady.bind(appController))
```
kirlat, line 189: We don't need this event any more. The only functionality that was related to it was a UI message. The message has been moved to the `HOMONYM_READY` handler , so the `MORP_DATA_READY` can be eliminated.

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1179,16 +1179,14 @@ export default class AppController {
     this._store.commit('app/lexicalRequestFinished')
   }
 
-  onMorphDataReady () {
-    this._store.commit('ui/addMessage', this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_READY'))
```
kirlat, line 1183: This message has been moved to the beginning of the `ON_HOMONYM_READY`. It is always published right after the `MORPH_DATA_READY`.

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1179,16 +1179,14 @@ export default class AppController {
...
-  }
-
   onMorphDataNotFound () {
     this._store.commit('ui/setNotification', { text: this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_NOTFOUND'), important: true })
     this._store.commit('app/setQueryStillActive', true)
   }
 
   onHomonymReady (homonym) {
+    this._store.commit('ui/addMessage', this.api.l10n.getMsg('TEXT_NOTICE_MORPHDATA_READY'))
```
kirlat, line 1188: This is where the message has been moved.

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -99,30 +109,24 @@ export default class LexicalQuery extends Query {
   }
 
   async getData () {
-    this.languageID = this.selector.languageID
-    if (this._hasLexemes) {
-      // Some lexical data is available, do a partial data retrieval
-      await this.partialDataRetrieval()
-    } else {
```
kirlat, line 106: This code was needed for the new lexical query workflow. All related functionality has been moved to `Lexis` now, and this path is no longer needed in the `LexicalQuery`.

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -362,87 +367,6 @@ export default class LexicalQuery extends Query {
     }
   }
 
-  async partialDataRetrieval () {
```
kirlat, line 365: This functionality has been split into three separate functions and was moved to the `Lexis`.

packages/components/src/lib/queries/lexical-query.js
```diff
@@ -473,27 +397,20 @@ export default class LexicalQuery extends Query {
       })
     }
     Query.destroy(this)
-    return result
   }
 
-  getLexiconOptions (lexiconKey) {
+  static getLexiconOptions ({ lexiconKey, languageCode, location, siteOptions, resourceOptions }) {
```
kirlat, line 402: This function has been made static to eliminate dependency on the `LexicalQuery` instance. This way it can be used from the `Lexis` module without creating an instance of the `LexicalQuery`.

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -32,8 +32,10 @@ export default class Lexis extends Module {
       Logger.getInstance().warn('CEDICT functionality will be disabled because CEDICT configuration is not available')
     }
 
+    // Whether lemma translations are enabled
+    this._lemmaTranslationEnabled = true
     // A locale for lemma translations (e.g. 'en-US')
-    this._lemmaTranslationLang = null
+    this._lemmaTranslationLocale = null
```
kirlat, line 38: Renamed to reflect the type of its value better.

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -405,10 +405,6 @@ export default class Lexis extends Module {
       return
     }
 
-    if (!lemmaTranslations && textSelector.languageID === Constants.LANG_LATIN && this._lemmaTranslationLang) {
-      // Use our own rules if lemmaTranslations is not provided
-      lemmaTranslations = { locale: this._lemmaTranslationLang }
-    }
```
kirlat, line 411: It is no longer required for the new workflow that do not depend on the `LexicalQuery`.

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -469,40 +465,62 @@ export default class Lexis extends Module {
...
       let annotatedHomonyms
       if (treebankWordIDs.length > 0) {
         annotatedHomonyms = await this.getTreebankData({
           store, textSelector, treebankDataItem
         })
       }
+
+      let lemmaTranslations
+      if (this._lemmaTranslationEnabled && textSelector.languageID === Constants.LANG_LATIN && this._lemmaTranslationLocale) {
+        // Use our own rules if lemmaTranslations is not provided
```
balmas, line 520: this comment doesn't make sense to me.

packages/components/tests/lib/queries/lexical-query.test.js
```diff
@@ -281,10 +281,8 @@ describe('lexical-query.test.js', () => {
   })
 
   it.skip('10 LexicalQuery - _getLexiconOptionsList parses lexicons', () => {
```
balmas, line 283: can we unskip this test now?

packages/components/tests/lib/queries/lexical-query.test.js
```diff
@@ -300,27 +298,27 @@ describe('lexical-query.test.js', () => {
...
+      location,
       siteOptions: allSiteOptions,
-      langOpts: {}
-    })
-    expect(query.getLexiconOptions('lexiconsShort')).toEqual({ allow: ['https://github.com/alpheios-project/xx'] })
+      resourceOptions: languageOptions
+    })).toEqual({ allow: ['https://github.com/alpheios-project/xx'] })
   })
 
   it.skip('11 LexicalQuery - _getLexiconOptionsList parses empty lexicons and returns {}', () => {
```
balmas, line 310: same here, can we unskip?

packages/components/src/vue/vuex-modules/data/lexis.js
```diff
@@ -523,6 +541,87 @@ export default class Lexis extends Module {
...
+    if (source !== LexicalQuery.sources.WORDLIST) {
+      LexicalQuery.evt.LEMMA_TRANSL_READY.pub(homonym)
+    } else {
+      LexicalQuery.evt.WORDLIST_UPDATE_LEMMA_TRANSL_READY.pub(homonym)
+    }
+    return homonym
+  }
+
+  // TODO: Should be moved behind the GraphQL facade
+  static async getWordUsageExamples ({ homonym, paginationAuthMax, clientId }) {
```
balmas, line 573: I think we should have an issue for further refactoring here. This method is only called if the `wordUsageExamplesON` setting was set to 'onLexicalQuery'. If it's set instead to 'onDemand', the still-existing LexicalQuery.getWordUsage data method is called instead. As we have discussed previously, I would like for all resource queries to be able to set as 'pre-cache' or 'on-demand' and for that to trigger when they are called, but we should not have two separate methods for calling them depending upon when they are called.  This is fine for this PR, but we do need to refactor it.

balmas (approved):
I think it's ok but see comments. Thanks!

irina060981 (approved):
It looks ok for me.
The only thing that I don't like - that before we had a huge UIController with a lot of types of data handling. Now we have a huge Lexis module with various methods of getting data. 
And it tightly connected with the other code by Vuex.Strore.

From my point of view - it would be better to move to the structure, where Lexis - is an independent module, with independent input/output. May be it is an independent package - that could be used with various applications (as we discussed on the meeting). And this way it could be divided to smaller classes that would be more independent, supportable and extendable.

packages/client-adapters/tests/adapters/tokenization/adapter.test.js
```diff
@@ -131,7 +132,7 @@ frigida pugnabant calidis, umentia siccis,
 mollia cum duris, sine pondere habentia pondus.
     `
     const result = await adapter.getTokens(text)
-    expect(result.segments.length).toEqual(20)
+    expect(result.segments.length).toEqual(2)
```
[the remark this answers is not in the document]
kirlat: Thank you!

packages/components/src/vue/vuex-modules/data/lexis.js
kirlat: Oh, you're right! The code being commented is executed only if the text selector's language is Latin. This code block is, however, in the `else` clause that will be executed only if the language is not Latin: https://github.com/alpheios-project/alpheios-core/blob/i566-resource-retrieval/packages/components/src/vue/vuex-modules/data/lexis.js#L434. As a result, the content of the `if` statement will never be run. Because of this, I've decided to remove it.

packages/components/tests/lib/queries/lexical-query.test.js
kirlat: Sure, enabled it now

packages/components/tests/lib/queries/lexical-query.test.js
kirlat: Has been enabled

packages/components/src/vue/vuex-modules/data/lexis.js
kirlat: Agree, I've created an issue to track this: #570.

kirlat:
> The only thing that I don't like - that before we had a huge UIController with a lot of types of data handling. Now we have a huge Lexis module with various methods of getting data.
> And it tightly connected with the other code by Vuex.Strore.
> 
> From my point of view - it would be better to move to the structure, where Lexis - is an independent module, with independent input/output. May be it is an independent package - that could be used with various applications (as we discussed on the meeting). And this way it could be divided to smaller classes that would be more independent, supportable and extendable.

Moving methods for resource retrieval to Lexis is a temporary solution. This is intended to leave the code in some workable state so I can switch to the Annotation UI and return to the resource retrieval after that. My intention is to move retrieval of lemma translations, word usage examples, and full definitions behind the GraphQL facade, similar to the short definitions. @balmas, @irina060981: would you think it would be a good idea?

From my point of view, all the logic to get all kinds of morphological data should be behind the GraphQL facade, and that might be separated into a module of its own. Whatever needs the morphological data should use the GraphQL to retrieve it. So the purpose of the Lexis module is to know how to retrieve the morphological data from GraphQL and construct in-memory objects (Homonyms and others) out of the GraphQL data. But there are also parts that cannot go behind the GraphQL facade, such as the Treebank data retrieval. It requires a document to be loaded within an iframe on each request and it can be done only from the application side. This functionality is in Lexis too.

I think that probably we will rip the greatest benefits now from (if not to count standardization of morphological data retrieval trough GraphQL) straightening the code that synchronizes events of data retrieval and UI updates (that's what the app controller does mostly). Probably it would make sense to put greater effort in there, to standardize interactions, to make code more testable, to eliminate functions that does nothing more than calling some other function (i.e. to shorten the call stack chains). But I think the ultimate priority should be the changes that adds value (i.e. that satisfy business requirements).

What do you think?

balmas:
I mostly agree. As we discussed on Monday, I think to truly have separation of concerns and more manageable code, we're going to have to introduce a session history so that we can keep track of the state of the application, and the data displayed in it at any moment, separately from the state of the queries.  Moving the queries behind the GraphQL facade will help, but it's not the whole story.

I also think the resource retrieval needs more thought before we move it behind GraphQL. We have some new requirements for resource retrieval coming up (see #545) that will need to be taken into account.

So I think as an intermediate step, these changes are fine, knowing it's temporary.

Issue #541, opened 2020-10-15
enable further lookups when no definition found


balmas:
Requested by @rgorman

Because of the date and subject matter of the text (and Marcus' own annoying predilection for obscure technical terms), many words are returned by Alph. as "no definitions found." In the nature of things, many of these are compounds.  E.G., πρό-ἀπολέγω. It would be great if we could click on either part of the lemma and be taken to the dictionary entry, since it is easy to guess the meaning of a compound from its parts, usually.

Similarly, Alph. sometimes says that the definition for some late term is the same as an earlier term (new = old, is the usual form). It would be great if we could call up the dictionary entry to the old form from this window.

balmas:
Another test case for this: ἐπεκυλίνδουν

balmas:
the fix to enable lookups on compound lemmas can be tested in Alpheios Components 3.3.1-qa.20210108437

lookups on greek words in definitions requires updated defs files, shoube tested via those issues.

monzug:
ἐπεκυλίνδουν   - 
1)  in the pop-up, can we highlight only the compound part that has been selected ( ἐπε    or   κυλίνδουν )  not the all word?

2) Would it be possible in this case to keep the original pop-up open so I can select also the other compound word (like nesting)? right now, when I click on ἐπε  , a new pop-up is generated and the original one is gone. if I want to lookup for the second word κυλίνδουν,  I have to enter the ἐπεκυλίνδουν word again.

balmas:
>Would it be possible in this case to keep the original pop-up open so I can select also the other compound word (like nesting)? right now, when I click on ἐπε , a new pop-up is generated and the original one is gone. if I want to lookup for the second word κυλίνδουν, I have to enter the ἐπεκυλίνδουν word again.

For this we really need session history, which will be part of the next major release

balmas:
>in the pop-up, can we highlight only the compound part that has been selected ( ἐπε or κυλίνδουν ) not the all word?

I think that change is too big for the incremental too unfortunately. The selection code is tricky. I'm going to enter a new issue for it to be considered for the future.

monzug:
will close this one as all questions have been addressed in new issues. Thanks

Issue #135, opened 2020-02-11
Mutiple Homonym from Morphology Request (Chinese)


irina060981:
For the feature - https://github.com/alpheios-project/client-adapters/issues/40

We need to add an ability 
   -  to get several homonyms from the lexicalRequest, 
   -  publish all of them to the popup (and may be panel - full definitions, but for now we have not that ability for Chinese) 
   -  and save all of the WordList.

We need to refactor:
  - LexicalQuery
  - UIController modules, Vuex parameters
  - WordListController
  - Popup/Panel components

I see the following steps to do here:

 -  as we have less space - I suggest to create tabbed popups, each tab would be a separate morph.vue component with current behaviour
 -  as for UIController modules and Vuex - I suggest to convert single homonym and other related attributes to objects (key = targetWord + LangugeCode, value = homonym) - may be we should create here a specific Class - like HomonymSet (in data-modules)?
     I suggest object here, because Map is not trackable bu VueComponents
 - convert LexicalQuery result to Homonym sets
 - update WordlistController to homonym sets

@balmas and @kirlat, what do you think?

balmas:
we need this feature also for treebank queries, where sometimes a single word in a source text is linked to multiple words in a treebank.

E.g. "virumque" is an enclytic that is a combination of the 2 words "virum" and "que", and in a treebank, these would likely be reflected in two separate tokens.

balmas:
Another example of compound words that needs to be supported from #541

For the greek word προἀπολέγω the morpheus parser returns 3 possible lexemes, each of which has a compound word as the lemma :

πρό,ἀπό-λέγω1  =>  3 separate words: πρό , ἀπό , and λέγω1
πρό-ἀπολέγω1 => 2 separate words πρό, and ἀπολέγω1
πρό-ἀπολέγω2 => 2 separate words πρό, and ἀπολέγω2

For a total of 5 distinct lemmas, each of which has an entry in our shortdefs file:

  "πρό|gen. before",
  "ἀπό|gen from",
  "λέγω1|say, speak",
  "ἀπολέγω1|to pick out from",
  "ἀπολέγω2|to decline, refuse"

Both our data model and the query needs to be able to account for this

For each lexeme, we should be able to show the user the separate components, with their definitions.

@kirlat this is something that needs to be handled properly in the lexical query refactoring and annotation code.

kirlat:
That's an interesting case to test with Thanks for bringing it in here!

balmas:
[...]

PR #575, opened 2020-12-11
fix for #541

balmas:
make lemma parts in the popup alpheios-enabled for lookup

This implements the request in #541  by enabling lookups on parts of compound lemmas in the popup.  The better solution would be to take the compound parts into account at the time of the lexicon lookup and include the definitions for all of the parts from the outset, but I would like to wait for the lexical query refactoring to do that, because it would require further hacking/tangling of the code to do it without it.  So #135 will address that, and in the meantime, this allows additional lookups by the user.

Note that the 2nd half of #541, which requests lookups on the definitions, is already technically possible in the code, but the greek short definitions index needs to be updated to mark the greek words in spans. That will be a data fix not a code fix.

Finally note that this PR merges into the `incr-3.3.x branch`, which is based off of the `production` branch. I need to do a small incremental release before the new semester and I think we have too much in master now to do it off that safely in time. So the incremental fixes will merge into `incr-3.3.x`, then to `qa` and `production` for release and back to master after release.


packages/components/src/vue/components/morph-parts/principal-parts.vue
```diff
@@ -6,13 +6,17 @@
         <h4
             class="alpheios-principal-parts__groupitem alpheios-principal-parts__groupitem--lemma-word"
             :lang="languageCode"
+            data-alpheios-enable="all"
             v-if="! lemma.principalParts.includes(lemma.word)"
-        >{{lemma.word}}</h4>
+        ><span
+          class="alpheios-principal-parts__groupitem--lemma-word__listitem"
+          v-for="word in lemma.word.split(/[,-]/)">{{word}}</span></h4>
```
kirlat, line 13: I think it might probably be better, rather than splitting in place, create a method on the Vue instance and handle it there. I expect after running it in the wild we might run into some not-so-trivial splitting cases that might require more complex conditions and error handling. Such logic better be separated into its own method, I think. Maybe it would even deserve a method in the `Lemma` class (but not now, of course). What do you think?

kirlat (approved):
Looks good, except for one comment, which might be opinionated though. It's amazing how such seemingly complex task was fulfilled with rather minor code changes!

packages/components/src/vue/components/morph-parts/principal-parts.vue
balmas: ok, you're probably right about making a method.

In terms of where it belongs in the long run, as I alluded to in #135, I think the query code needs handle this and create a data model object that more accurately reflects what this is -- i.e a compound word with multiple lemmas.

PR #576, opened 2020-12-11
more for #541

balmas:
activate text selector on short definitions always (only has effect on alpheios-enabled elements)
make sure styling applies to sub elements of the short definition
make the cursor a pointer on any alpheios-enabled elements


kirlat (approved):
Looks good!

PR #577, opened 2020-12-11
fix #55

### LINKED ISSUE #55: missing the tense and voice in Greek participle verbs (and Latin verbs too)
lookup  κατασκευάσας     - see screenshot below

verb-participle

@balmas: I think because it has a case, which is overriding the voice and tense information.

Comparing with the latin verbs (e.g. reputantis), I realized that tense and voice are also missing from the latin morphology pop-up but we do show the conjugation which we don't have in Greek verbs and in Morpheus. to add more info, should we consider the stemtype?

balmas:
this just removes the restriction on displaying tense and voice (if they're present in the inflection data) in the popup for #55 

(that this PR merges into the incr-3.3.x branch, which is based off of the production branch. I need to do a small incremental release before the new semester and I think we have too much in master now to do it off that safely in time. So the incremental fixes will merge into incr-3.3.x, then to qa and production for release and back to master after release.)


irina060981 (approved):
Looks ok

PR #582, opened 2020-12-16
First draft of an Annotation UI

kirlat:
Here is the first draft of the Annotation UI. The code is not optimized yet: it's too early for that. The main purpose is to introduce the framework on which the annotation components be built and to offer some mockup of the UI for discussion.

There are no tests yet as there are not real functionality in the components, but they will be added with the new functionality.

Please let me know what do you think. Any suggestions (especially about the architecture) are very welcome!


irina060981 (commented):
I didn't finally understand:
- why do you have styles for annotations UI both inside annotations package and components package
- what is the input point of the anotations code inside components

irina060981:
I uploaded your branch and looked at the UI prototype.
As I was not inside any discussions how they would be useful for the end user's - I couldn't estimate how they are compatible with user's expectations.
But they look rather nice 👍 

I saw only a problem - that popup became too longer (more then height of the screen), after I closed blocks and turned off annotations, the height of the popup stayed the same. And I was not able to move it properly or change the size.

packages/annotations/src/ui-components/area.vue
```diff
@@ -0,0 +1,112 @@
...
+  props: {
+    annotationMode: {
+      type: Boolean,
+      required: false,
+      default: false
+    }
+  },
+
+  methods: {
+    clickHandler: function (evt) {
```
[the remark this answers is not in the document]
kirlat: Oh, this is a good catch! I was somehow not paying any attention to this. I think I can use `<function name> (<arguments>) {` everywhere as it is an equivalent of `<function name>: function (<arguments>) {` but shorter.

packages/annotations/src/ui-components/area.vue
```diff
@@ -0,0 +1,112 @@
...
+      const target = evt.target.closest(`[${ANNOTATABLE_ATTR_NAME}]`)
```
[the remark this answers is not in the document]
kirlat: This attribute serves as an indicator of what elements would be allowed to add/edit/remove annotations. Let's take. for example, https://github.com/alpheios-project/alpheios-core/blob/dev-annotation-ui-next/packages/components/src/vue/components/morph-parts/principal-parts.vue#L14. The presence of the `data-annotation-type` is an indication that this element must have annotations related functionality. The value, `comment` specifies what type of annotation functionality should be available.

My goal is to make the annotations UI as unobtrusive as possible. The source code where annotation functionality to be added should require as little modifications as possible. That's why the use of data attributes. Thy do not require components that use them to import anything. Only one component is necessary: the root `AnnotationArea`  component that should wrap the code with the `data-annotation` tags. See https://github.com/alpheios-project/alpheios-core/blob/dev-annotation-ui-next/packages/components/src/vue/components/morph.vue#L3 as an example. 

This component, once activated, will listen for the clicks within its inner content. Then, if a click comes from an element whose parent has a `data-annotation-type` attribute, it assumes the element is in the edit mode (or switches from the edit mode to the normal one), and it creates a Vue component that represents an editing panel, and inserts it below the element. Annotation package is also responsible for bringing in the CSS for the annotatable elements.

The only part I'm not quite happy with are panels to add new item types (lemma, inflections, etc.). Right now they require corresponding annotation components to be imported, as in https://github.com/alpheios-project/alpheios-core/blob/dev-annotation-ui-next/packages/components/src/vue/components/morph.vue#L23.

I think ideally there should be the code like `data-annotationtype="lemma"` and, when annotation root component is mounted, it should be replaced with the component's code. The problem here, however, is that the `AnnotationArea` component is mounted before the `morph` one and thus it is too early for the `AnnotationArea` to access to HTML code of the `morph`'s template. Not sure how deal with it. Maybe to have one generic component for adding panels like this that will insert the corresponding panel inside itself. This would eliminate the need of the client to import different kinds of components. It is enough for it to know and import only one component type.

Please let me know if this makes sense.

kirlat:
> * why do you have styles for annotations UI both inside annotations package and components package

This is I overlooked, it was in the `components` before but I've forgotten to remove it. Will fix~ 

> * what is the input point of the annotations code inside components

Please see an explanation in https://github.com/alpheios-project/alpheios-core/pull/582#discussion_r545061205.

kirlat:
> I uploaded your branch and looked at the UI prototype.
> As I was not inside any discussions how they would be useful for the end user's - I couldn't estimate how they are compatible with user's expectations.
> But they look rather nice 👍

Thanks! But I guess it will come through many iterations before the final design will be born 🙂.
> 
> I saw only a problem - that popup became too longer (more then height of the screen), after I closed blocks and turned off annotations, the height of the popup stayed the same. And I was not able to move it properly or change the size.

That's true, but I think probably we should postpone fixing this before other changes will be implemented. Next we will need to add comments that can take a lot of space so they will expand the popup even more. They may also be shown in a panel as an alternative. Not sure what the final design will be, but we definitely need some way to adjust popup dimensions for the annotation mode.

packages/annotations/src/ui-components/area.vue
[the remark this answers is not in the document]
balmas: I like the approach. I think it's an interesting question where the knowledge about which component to import should go.  On the one hand, it's appealing to have that all encapsulated in the annotation package, but on the other hand, the domain knowledge about how a specific data object should be annotated probably belongs with the domain component.   I think this will become clearer as we design the graphQL api for the mutations.

packages/annotations/src/ui-components/comment.vue
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export default {
+  name: 'AnnotationComment',
+  data () {
+    return {
+      selectedAction: ActionTypes.NONE
+    }
+  },
+  computed: {
```
balmas, line 55: Maybe we should have a base component that has all of the basic methods (isAdding, isRemoving, isFormOpen) that the individual components can be derived from? This way we could make sure the functionality that is common is stable across all and avoid code duplication?

packages/annotations/src/ui-components/comment.vue
```diff
@@ -0,0 +1,86 @@
...
+        <div class="alpheios-annotations__act-form-content">
+          <div class="alpheios-annotations__act-form-headline">Edit the comment:</div>
+          <div class="alpheios-annotations__act-form-text-input-group">
+            <textarea id="lemma-definition" cols="50"
+                      name="lemma-definition" rows="4">A comment about the part of speech of the lexeme</textarea>
+          </div>
+        </div>
+
+        <div class="alpheios-annotations__act-form-ctrls">
+          <div class="alpheios-annotations__act-form-ctrls-save" @click.stop="closeForm">Save</div>
```
balmas, line 21: I know it's just a prototype but I think I would still like for all of the text to be in messages files (Save, Remove, Add, Cancel, etc.)

packages/annotations/src/ui-components/inflection.vue
```diff
@@ -0,0 +1,84 @@
...
+      </div>
+
+      <div class="alpheios-annotations__act-form" v-show="isAdding">
+        <div class="alpheios-annotations__act-form-content">
+          <div class="alpheios-annotations__act-form-headline">Add an inflection:</div>
+          <div class="alpheios-annotations__act-form-dropdown-group">
+            <label for="latin-number">Number:</label>
+            <select name="latin-number" id="latin-number">
+              <option value="not-selected">Select a value</option>
+              <option value="Singular">Singular</option>
```
balmas, line 16: Just a note for later - - we are going to have to have the domain component tell us what the options are -- they will change depending on the language, particular word, etc.

packages/components/src/vue/components/morph-parts/inflections-list.vue
```diff
@@ -4,14 +4,15 @@
...
             <span class="alpheios-inflections-list__inflset_index" v-if="inflections.length > 1">{{ ifindex + 1 }}.</span>
 
-            <div class="alpheios-inflections-list__forms">
+      <div class="alpheios-inflections-list__forms">
                 <span
                     v-for="feat in featuresList.wordParts.filter(feat => inflset.groupingKey[feat.name])" v-bind:key="feat.name"
                     :lang="languageCode" class="alpheios-inflections-list__formtext"
                     data-grouplevel="1" data-feature="feat.name"
                 >{{ feat.template.replace('%s', inflset.groupingKey[feat.name]) }}</span>
+        <annotation-word-form></annotation-word-form>
```
balmas, line 13: So we can't use the data-annotation-type attribute here because the annotation component is drawn first?

balmas (commented):
I think it's a good start . A few small questions and requests.

packages/components/src/vue/components/morph-parts/inflections-list.vue
kirlat: Yes, since the `AnnotationArea` is the outer component, it is mounted first. Once it's loaded, Vue loads inner components, such as `AnnotationWordForm`. Because of this, to instantiate an `AnnotationWordForm` from the  `AnnotationArea.mount()` would be too early. But maybe there is some other way to handle this.

packages/annotations/src/ui-components/inflection.vue
kirlat: That's a tricky problem we have to solve. The super important questions are where does this knowledge should be and how the annotations would get access to it.

packages/annotations/src/ui-components/comment.vue
kirlat: That's a good idea. will add that.

packages/annotations/src/ui-components/comment.vue
kirlat: I was thinking about that too, but I decided to postpone this for later. Right now components are simple because they do almost nothing. I was afraid that once we start to put more complex logic into them, they would start to drift away from each other. But I agree the common base class is a great idea. Will add that.

balmas: ok. I'm fine with it being postponed to another PR

kirlat:
Will merge it at its current state. Anything from the suggestions that has not been implemented will be implemented in the next PR(s).

Issue #587, opened 2020-12-21
tokenization adapter misses the 'ignore' option for the tei endpoint


balmas:
reported by @monzug 

The 'ignore' option if supplied in the parameters to the tokenize adapter fetchOptions does not get passed on to the service.

I think it would be better if the adapter did not hard code the option names, but instead retrieved them from the service schema.

monzug:
similar behavior with Supply a comma-separated list of elements in the text that represent visual blocks
I added "l,p" to the list and compare with an empty list, (removed all the elements), the line breaks in the alignment looked the same. I did also try with 'div". the origin and target text had l,p and div tags.

irina060981:
Merged and closed

monzug:
Supply a comma-separated list of elements in the text that represent visual blocks works now. tested with l, p and div.

Issue #141, opened 2018-10-28
Remove numbers from morphology display


monzug:
@balmas said: This one is a little tricky.  The numbers allow us to find the right 
definition in lsj when there are multiple entries for a homonym, because 
the morpheus lemmas are directly drawn from the lsj entries.  They are 
less helpful for the other dictionaries though.

I think it needs to be addressed at some point, but will probably not do 
it in a patch release because I would be afraid of breaking things.
![uno-due](https://user-images.githubusercontent.com/41396793/47617953-9474b100-daa3-11e8-9912-cce647c089a0.JPG)
move number from 

https://github.com/alpheios-project/alpheios-core/issues/198

οὐδ'

irina060981:
I believe that these issues are about the same theme -
https://github.com/alpheios-project/alpheios-core/issues/198

balmas:
yes, related but I think #198 is a non-issue, as shown by the example in this issue.

So here's the situation:

when a lemma is a homonym with multiple lexemes with different meanings sometimes the morphology service (morpheus parser) returns lemmas with a trailing digit to distinguish them.

The lemmas are indexed with these in the short definitions files, so if we get, for example,  οὐδός2, as the lemma from the parser, we can correctly retrieve "way" as the definition for that lexeme, rather than "threshold"

We also have a fallback in the lexicon adapter code, which tells the lookup to try without the trailing digit if the lemma isn't found with it in the index, because these references are specific to the lsj lexicon and likely won't be in other lexical sources.  (e.g. οὐδός could just be indexed once without distinguishing the different lexemes). 

So it's important to keep the digit in the data that is used for looking up in the lexicons, but it not doesn't add any value to the display and it could prevent matching in inflection tables, etc.

irina060981:
Merged and closed

balmas:
I'm going to port this to the incremental release

balmas:
fixed in Alpheios Components 3.3.1-qa.20210108437

monzug:
fixed for οὐδ' in build   Alpheios Reading Tools 3.3.1 build qa.20210108447
Alpheios Components 3.3.1-qa.20210108437 

Unfortunately I could not find any other example to test.

![no-number-greek-popup](https://user-images.githubusercontent.com/41396793/104041182-a6376000-51a6-11eb-8275-64bee75ed439.JPG)

balmas:
[...]

PR #588, opened 2020-12-23
Remove numbers from lemmas (i141) and Fix tokenization service error (i587)

irina060981:
For the #141 
- added new method to the Lemma - `displayWord`, that removes numbers
- print to popup `displayWord` instead of word
- added a small unit test

For the #587, https://github.com/alpheios-project/alignment-editor-new/issues/78
- updated `createTokenizeFetchURL` to be used with non predefined options
- added a small unit test


packages/components/src/vue/components/morph-parts/principal-parts.vue
```diff
@@ -7,7 +7,7 @@
             class="alpheios-principal-parts__groupitem alpheios-principal-parts__groupitem--lemma-word"
             :lang="languageCode"
             v-if="! lemma.principalParts.includes(lemma.word)"
```
balmas, line 9: this one is tricky and is likely going to need to be revisited with the annotation work of creating identifiers for lexical entities.  I think it's okay to include it, but probably we also need to use the displayWord when checking whether the word appears in the principalParts on line 9

balmas (changes requested):
Thanks.

packages/components/src/vue/components/morph-parts/principal-parts.vue
irina060981: I am not quite sure about it, because I don't know the example - with numbered lemmas that has also principal parts.
So I don't know what word's variant would be there - but I believe from your notice that it won't have numbers?
I fixed according to your comment, as you are more well-known here :)

irina060981:
@balmas , I think I could merge this PR as I fixed your only one note.

PR #592, opened 2020-12-28
Initial support for the IRIs

kirlat:
This build introduces the new IRI provider class. Its task is to generate unique IDs. So far it can generate two types: content based (two items with the same content will have the same ID) and unique (every item has it's own unique IDs). 

Only definitions class has been updated to use the IRI provider so far, but if the concept is workable, we can switch all our classes to the IRI provider, with one or the other IRI type.

In order to generate a content based IRI, a fields that represent the content should be designated somehow. This is what the `identityData` getter is for: it returns props that participate in generation of the content-based IDs. Each class that need content based IDs should have a method like that.

I also think that it is important to "protect" fields that participate in the content-based ID generation from the change: if they are altered the ID would lose its actuality and will need to be regenerated. So those fields are declared as "private" (it does not provide any protection technically, but it's better then nothing; probably real private fields will be in JS one day) and are accessed through getters.

There is also another set of changes that switches some existing classes to use the Language class instead on language ID.


packages/client-adapters/src/adapters/chineseloc/adapter.js
```diff
@@ -38,6 +38,8 @@ class AlpheiosChineseLocAdapter extends BaseAdapter {
     this._messagingService = MessagingService.getService(msgServiceName)
   }
 
+  get language () { return ChineseLanguageModel.language }
+
   get languageID () { return ChineseLanguageModel.languageID }
```
balmas, line 43: do we still need this method?

packages/client-adapters/src/adapters/tufts/engine/sedra.js
```diff
@@ -11,9 +11,9 @@ data.setLexemeFilter(function (lexeme) {
 
 //
 data.setMeaningParser(function (meaning, targetWord) {
-  const lang = meaning.lang ? meaning.lang : Models.Constants.STR_LANG_CODE_ENG
+  const lang = meaning.lang ? new Models.Language(meaning.lang) : Models.Language.ENGLISH
   const meaningText = meaning.$ || ''
-  return new Models.Definition(meaningText, lang, 'text/html', targetWord)
+  return new Models.Definition(meaningText, lang, Models.Constants.MIMETypes.TEXT_PLAIN, targetWord)
```
balmas, line 16: this should be TEXT_HTML

packages/client-adapters/tests/adapters/lexicons/adapter.test.js
```diff
@@ -42,6 +42,7 @@ describe('lexicons/adapter.test.js', () => {
   beforeEach(async () => {
     ClientAdapters.init()
     testLangID = Constants.LANG_GREEK
```
balmas, line 44: do we need testLangID anymore?

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -878,16 +878,28 @@ export default class AppController {
   /**
    * Gets language name details by either language ID (a symbol) or language code (string)
    *
-   * @param {symbol|string} language - Either language ID or language code (see constants in `data-models` for definitions)
+   * @param {Language | symbol|string} language - A language for which a name should be retrieved.
```
balmas, line 881: should we note that symbol and string are deprecated value types?

packages/components/tests/vue/components/word-usage-examples/word-usage-examples.test.js
```diff
@@ -27,7 +27,10 @@ describe('word-usage-examples.test.js', () => {
     testWordUsageList = await BaseTestHelp.collectConcordance(homonym)
   })
 
-  beforeEach(() => {
+  beforeEach/**
```
balmas, line 30: maybe an ide error here with the inserted comments?

packages/data-models/src/constants.js
```diff
@@ -234,3 +234,11 @@ export const CLASS_RECIPROCAL = 'reciprocal'
 // Paradigms
 export const PARADIGM_CAT_KAYLO = 'kaylo'
 export const PARADIGM_CAT_STATE = 'state'
+
+/**
+ * Constants that define MIME Types used in Alpheios.
+ *
+ * @enum {string} */
+export const MIMETypes = {
+  TEXT_PLAIN: 'text/plain'
```
balmas, line 243: need also text/html

packages/data-models/src/definition-set.js
```diff
@@ -1,25 +1,56 @@
-import Definition from './definition'
+import Definition from './definition.js'
+import Language from './language.js'
 import LMF from './language_model_factory.js'
 
 export default class DefinitionSet {
-  constructor (lemmaWord, languageID) {
+  /**
+   * @param {string} lemmaWord - A word for which a definition set was created.
+   * @param {Language} language - A language on what a text of a definitions in a definition set is written.
```
balmas, line 8: I believe the language property on the definition set is the language of the lemma and not the definition text itself.

packages/data-models/src/definition.js
```diff
@@ -1,22 +1,90 @@
...
+  }
+
+  get ID () {
+    return this._ID
+  }
+
+  get identityData () {
+    return {
+      text: this._text,
+      languageCode: this._language.toCode()
```
balmas, line 77: I think we should also include the resource provider uri in the identity data for the definition to start. There may be additional changes to the way we calculate the IRI for definitions down the road, but that would be a reasonable place to start.

packages/data-models/src/homonym.js
```diff
@@ -72,21 +72,29 @@ class Homonym {
...
+      lexemes = [new Lexeme(new Lemma(targetWord, lang), [])]
     }
-    const homonym = new Homonym(lexemes, jsonObject.form || jsonObject.targetWord)
+    const homonym = new Homonym(lexemes, targetWord)
     homonym.lemmasList = jsonObject.lemmasList
     return homonym
   }
 
   convertToJSONObject (addMeaning = false) {
+    // TODO: Shall we use targetWord instead of form to match the internal homonym structure?
```
balmas, line 97: we probably should. but let's create a separate issue to track that because it could impact the wordlist.

packages/data-models/src/iri/iri-provider.js
```diff
@@ -0,0 +1,96 @@
...
+        return IRIProvider._getUUIDv4()
+      }
+    } else if (type === IRIProvider.IRITypes.MD5_HASH) {
+      if (!IRIProvider._isValidIdentityData(identityData)) {
+        throw new Error(IRIProvider.errMsgs.NO_IDENTITY_DATA)
+      }
+      return IRIProvider._getMD5Hash(identityData)
+    } else if (type === IRIProvider.IRITypes.UUID_V4) {
+      return IRIProvider._getUUIDv4()
+    }
```
balmas, line 36: else throw error?

packages/data-models/src/iri/iri-provider.js
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Checks wither the identity data is valid.
+   * The object is valid if it contain at least one key-value pair.
+   *
+   * @param {object} identityData - An identity data object.
+   * @returns {boolean} - True if the object is valid, false otherwise.
+   * @private
+   */
+  static _isValidIdentityData (identityData = {}) {
+    return Boolean(identityData && Object.keys(identityData).length > 0)
```
balmas, line 48: maybe we should also test that the values of the identityData hash are populated with strings?

packages/data-models/src/language.js
```diff
@@ -19,18 +20,26 @@ export default class Language {
   /**
    * Creates an instance of a language class.
    *
-   * @param {Lang} code - A constant that specifies a language.
+   * @param {Lang | string} code - A constant that specifies a language in an ISO 639-3 format.
```
balmas, line 23: Since we are now using the Language class to encapsulate languages other than those that initiated their creation (e.g. for a Definition, we have the language of the lemma of the definition and the language of the text of the definition) we should explicitly support ISO 639-2 codes as well as ISO 639-3. For Alpheios-supported languages we normalize to ISO 639-3 but for other languages we should support either, and we could validate against the standard list of 639-3 and 639-2 codes.

packages/data-models/src/language_model_factory.js
```diff
@@ -27,12 +27,22 @@ class LanguageModelFactory {
   /**
    * Checks whether a language is supported
    *
-   * @param {string | symbol} language - Language as a language ID (symbol) or a language code (string)
+   * @param {Language | string | symbol} language - Language as a Language object,
```
balmas, line 30: should we note that string and symbol are deprecated?

balmas (commented):
(github is acting a little weird so I'm going to submit these comments to make sure they don't get lost. I'm not fully through the review yet).

packages/data-models/src/lemma.js
```diff
@@ -73,7 +92,7 @@ class Lemma {
     // eslint-disable-next-line prefer-const
     let resultLemma = {
       word: this.word,
-      language: this.languageCode,
+      languageCode: this.language.toCode(),
```
balmas, line 95: note that this change could affect the wordlist.

packages/inflection-tables/src/paradigm/data/greek/greek-paradigm-dataset.js
```diff
@@ -24,6 +24,10 @@ export default class GreekParadigmDataset extends LanguageDataset {
       .map('future_perfect', [Constants.TENSE_FUTURE_PERFECT])
   }
 
+  static get language () {
+    return Language.GREEK
+  }
+
   static get languageID () {
```
balmas, line 31: do we need languageID getter now?

packages/inflection-tables/lib/lang/latin/latin-language-dataset.js
```diff
@@ -71,6 +71,10 @@ export default class LatinLanguageDataset extends LanguageDataset {
     }
   }
 
+  static get language () {
+    return Language.LATIN
+  }
+
   static get languageID () {
```
balmas, line 78: do we need languageID getter now?

packages/data-models/src/language.js
```diff
@@ -9,7 +9,8 @@ export const Lang = {
   PERSIAN: 'per',
   GEEZ: 'gez',
   CHINESE: 'zho',
-  SYRIAC: 'syr'
+  SYRIAC: 'syr',
+  ENGLISH: 'eng'
```
balmas, line 13: I don't actually know if we should normalize 'en' to 'eng'  Was this added for the definition languages? if so, then see other comment on this topic.

balmas, line 13: I think we still have more cleanup to do here with the language codes of the alpheios supported languages --- right now this list has to be updated in too many places when we add support for a new source language. (see https://github.com/alpheios-project/documentation/blob/master/development/adding_a_language.md which is almost certainly out of date). I'm not sure what the best solution is and I don't think it has to be solved with this PR but we should keep sight of it.

balmas (approved):
for the most part, this looks okay to me. I have a number of small questions and comments but on the whole it looks good.  I suspect we have broken things (possibly with the wordlist) but hopefully we can work through those as we find them and continue with the cleanup efforts.

packages/data-models/src/definition.js
```diff
@@ -30,11 +98,11 @@ class Definition {
...
     // eslint-disable-next-line prefer-const
     let result = {
-      text: this.text,
-      language: this.language,
+      text: this._text,
+      languageCode: this._language.toCode(),
       format: this.format,
       lemmaText: this.lemmaText,
-      ID: this.ID
+      ID: this._ID
```
irina060981, line 105: I don't really understand the sense of dividing properties to public and private on this stage.
I believe (it is only my thoughts), that one of the reasons to divide propetries to public and private, that
- public could be needed directly in other classes
- private could not be used directly

But you create methods with making `_`like properties public and they could be used directly anyway with additional lines of code.

Another reason to create private properties - is to scope changes of such properties inside one method and don't allow to change it directly. But you change such properties directly (like `_ID`).

Is this additional code is really useful and give us any benefits?

packages/data-models/src/homonym.js
```diff
@@ -95,32 +103,29 @@ class Homonym {
...
-   * by using language property of the first lemma. We chan change this logic in the future if we'll need to.
+   * by using language property of the first lemma. We can change this logic in the future if we'll need to.
    *
-   * @returns {string} A language code, as defined in the `languages` object.
+   * @returns {Language} A language of the homonym.
    */
   get language () {
-    Logger.getInstance().warn('Please use languageID instead')
-    return LMF.getLanguageCodeFromId(this.languageID)
+    if (this.lexemes && this.lexemes[0] && this.lexemes[0].lemma && this.lexemes[0].lemma.language) {
```
irina060981, line 114: May be it is worth to create a specific check method inside Lexeme class?

packages/data-models/src/iri/iri-provider.js
irina060981: It is only a suggestion :)
I don't like long if/else if/else constructions - they are not easy to read, and it is a "headache" to expand them.
May be it is better to create a simple const list like
```
 IRIType: { method: getIRIMethod, needCheckForIdentity: {Boolean} }
 ```
 
 We use such lists in different places in our code - and it is rather useful.
 @kirlat , what do you think?

**Review by irina060981 — APPROVED** [review]
I think that it is really difficult to update to usage of Languge class.
And I like how you organize it in general.
I have only some small suggestions.

I believe that IRI IDs should work well.

I have only worries with Translations and Wordlist.
But you told on the meeting that they are tested - so I believe it is allright :)

File: packages/client-adapters/src/adapters/chineseloc/adapter.js (hunk 1 above)
Comment (kirlat) [reply packages/client-adapters/src/adapters/chineseloc/adapter.js:43]: Unfortunately, `Feature` and some other classes still use language ID, so I think we have to have it here. Here is one of the usages of the language ID within the Chinese local adapter: https://github.com/alpheios-project/alpheios-core/blob/dev-annotations-next/packages/client-adapters/src/adapters/chineseloc/adapter.js#L166
But I think we can get rid of it some day.

File: packages/client-adapters/src/adapters/tufts/engine/sedra.js (hunk 2 above)
Comment (kirlat) [reply packages/client-adapters/src/adapters/tufts/engine/sedra.js:16]: Good catch, thank you! Will fix this.

File: packages/client-adapters/tests/adapters/lexicons/adapter.test.js (hunk 3 above)
Comment (kirlat) [reply packages/client-adapters/tests/adapters/lexicons/adapter.test.js:44]: Yes, unfortunately. Client Adapters require a language ID as a parameter, so the language ID must be here: https://github.com/alpheios-project/alpheios-core/blob/dev-annotations-next/packages/client-adapters/tests/adapters/lexicons/adapter.test.js#L53

File: packages/components/src/lib/controllers/app-controller.js (hunk 4 above)
Comment (kirlat) [reply packages/components/src/lib/controllers/app-controller.js:881]: That's a good idea, I will add that.

File: packages/data-models/src/definition.js (hunk 18 above)
Comment (balmas) [reply packages/data-models/src/definition.js:105]: Even though javascript doesn't enforce public and private class members, I do think it's helpful to have code standards for identifying the intent because it makes it easy to see when it is being accessed from outside the intended scope. I think using the _ for this is pretty standard and well understood.

File: packages/components/tests/vue/components/word-usage-examples/word-usage-examples.test.js (hunk 5 above)
Comment (kirlat) [reply packages/components/tests/vue/components/word-usage-examples/word-usage-examples.test.js:30]: Yes, I think it's a result of a glitch. Removed it.

File: packages/data-models/src/constants.js (hunk 6 above)
Comment (kirlat) [reply packages/data-models/src/constants.js:243]: Added that.

File: packages/data-models/src/definition-set.js (hunk 7 above)
Comment (kirlat) [reply packages/data-models/src/definition-set.js:8]: Oh yes, that's correct! Fixed the description.

File: packages/data-models/src/definition.js (hunk 8 above)
Comment (kirlat) [reply packages/data-models/src/definition.js:77]: It would be fixed in a separate PR because the complicated nature of the solution would make this PR too heavy. I've created an issue to track the progress: https://github.com/alpheios-project/alpheios-core/issues/595.

File: packages/data-models/src/homonym.js (hunk 9 above)
Comment (kirlat) [reply packages/data-models/src/homonym.js:97]: I've created https://github.com/alpheios-project/alpheios-core/issues/596 to track this.

File: packages/data-models/src/iri/iri-provider.js (hunk 10 above)
Comment (kirlat) [reply packages/data-models/src/iri/iri-provider.js:36]: Definitely, we should do that. I've updated the code.

File: packages/data-models/src/iri/iri-provider.js (hunk 11 above)
Comment (kirlat) [reply packages/data-models/src/iri/iri-provider.js:48]: Great idea, will add that check.

File: packages/data-models/src/language.js (hunk 12 above)
Comment (kirlat) [reply packages/data-models/src/language.js:23]: It might be a work with many nuances. I've created an issue https://github.com/alpheios-project/alpheios-core/issues/597 to track that.

File: packages/data-models/src/language_model_factory.js (hunk 13 above)
Comment (kirlat) [reply packages/data-models/src/language_model_factory.js:30]: Absolutely! I've add that note.

File: packages/data-models/src/lemma.js (hunk 14 above)
Comment (kirlat) [reply packages/data-models/src/lemma.js:95]: I've checked usages in the wordlist and tried to fix whatever may cause problems, run unit tests, and tested some wordlist cases manually. That definitely might be not enough, but I hope if there are any issues, they will be caught during Monica's full scale test. Tried to do my best not to break anything here.

File: packages/inflection-tables/src/paradigm/data/greek/greek-paradigm-dataset.js (hunk 15 above)
Comment (kirlat) [reply packages/inflection-tables/src/paradigm/data/greek/greek-paradigm-dataset.js:31]: It's used in couple of places in the constructor (https://github.com/alpheios-project/alpheios-core/blob/dev-annotations-next/packages/inflection-tables/src/paradigm/data/greek/greek-paradigm-dataset.js#L15-L16), but we could possibly be replaced with a constant. The problem with `languageID` is that it's used all over the place in the inflection tables and it's hard to track those usages exactly. I've tried to reduce changes to inflection tables to the minimal amount as changes of this library are out of scope for the current annotation work. So maybe we'll leave it as is for now? We can later add a log statement to track its usage.

File: packages/inflection-tables/lib/lang/latin/latin-language-dataset.js (hunk 16 above)
Comment (kirlat) [reply packages/inflection-tables/lib/lang/latin/latin-language-dataset.js:78]: Same as above: I'm not sure, but maybe it's safer to leave as is for now?

File: packages/data-models/src/language.js (hunk 17 above)
Comment (kirlat) [reply packages/data-models/src/language.js:13]: I agree it may be not a clear cut. I think we can discuss it in the corresponding issue https://github.com/alpheios-project/alpheios-core/issues/597 as this might get nuanced.

File: packages/components/src/vue/components/morph-parts/definitions-list.vue [hunk 20]
```diff
@@ -44,7 +44,12 @@ export default {
       if (this.$store.getters['app/shortDefDataReady'] && this.lexeme && this.lexeme.meaning && this.lexeme.meaning.shortDefs && this.lexeme.meaning.shortDefs.length > 0) {
         definitionsLocal = this.lexeme.meaning.shortDefs
       } else if (this.lexeme && this.lexeme.lemma.features && Object.entries(this.lexeme.lemma.features).length > 0) {
-        definitionsLocal = [new Definition(this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), 'en-US', 'text/plain', this.lexeme.lemma.word)]
+        definitionsLocal = [
+          new Definition(
+            this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), Language.ENGLISH,
```
irina060981, line 49: @kirlat , some comments were erased somehow - I would repeat with this ine - because it is a error.
You pass `Language.English` instead of 'en-US' String, I believe `getMsg` method does not support passing Object.
You should fix this, I think.

packages/data-models/src/definition.js
irina060981: Then if we decide to use this as a code standard for all outr team - let's define - what class properties should be defined as private and what should be defined as public. 

What are exactly conditions to define a class propety as private? I suggest to define them and print somewhere :)

packages/data-models/src/language.js
kirlat: We could probably move the discussion to the https://github.com/alpheios-project/alpheios-core/issues/597 and handle this issue separately because it might require some analysis in order to find the best solution.

packages/components/src/vue/components/panel-compact.vue
```diff
@@ -493,7 +493,12 @@ export default {
           if (lexeme.meaning.shortDefs.length > 0) {
             definitions.push(...lexeme.meaning.shortDefs)
           } else if (Object.entries(lexeme.lemma.features).length > 0) {
-            definitions.push(new Definition(this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), 'en-US', 'text/plain', lexeme.lemma.word))
+            definitions.push(
+              new Definition(
+                this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), Language.ENGLISH,
+                Constants.MIMETypes.TEXT_PLAIN, lexeme.lemma.word
```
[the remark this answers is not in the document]
kirlat: Yes, those two are different, but I believe I was fixing a mistake that we made before. Here we create an instance of a `Definition` class, and the second parameter of its constructor is the language on what the text of the definition is written.

Previously, the `Definition` constructor was taking a language code. I've inspected the usages and in absolute majority of cases it was `eng` with `en` being the rest. `en-US` is the locale. Locale, to the best of my understanding, expresses user preferences related to the location, such as date, time formats, and may, or may not, depending on the implementation, control the language on which messages are written (for example, in MS Visual Studio locale and language are two different concepts: https://github.com/alpheios-project/alpheios-core/issues/597). 

The `Definition`, especially since its parameter was switched to be the `Language` object, is clearly requiring a language, not a locale. So the `Language.ENGLISH` (with the language code `eng`), not the `en-US` locale, is the most appropriate argument, on my opinion.

This, however, rises an issue I didn't realize before: in cases like that we need not to hardcode the language, but take it from the `l10n` instance (what is the language of the message that was just returned?). So maybe we should modify the `L10n` class to return the language of the currently selected console. 

@balmas, @irina060981: What do you think?

packages/data-models/src/definition.js
kirlat: With content-based IDs most of our objects that have content-based IDs should be the value objects (in terms of Domain Driven Design): they should represent a value, or a set of values, and if the value(s) is(are) changed, we should create another instance of the object with new values rather than mutate values within the current object.

This code follows this approach, I believe. The values that affect the ID, and the ID values itself should be immutable. I think the best way to achieve this would be to make the corresponding props private, accessible only from within the class, (well, it does not provide any protection now but the support of "real" private fields is already in Chrome, Edge, and FF with the flag) and provide getters only. This would make the values read only: no one outside the class will be able to mutate their values. So that was the idea.

@balmas, @irina060981: what do you think? Does it make sense?

packages/data-models/src/homonym.js
kirlat: I think it's a very good proposal, I will create a method for the check.

kirlat:
> I don't like long if/else if/else constructions - they are not easy to read, and it is a "headache" to expand them.
  `IRIType: { method: getIRIMethod, needCheckForIdentity: {Boolean} }`
We use such lists in different places in our code - and it is rather useful.
@kirlat , what do you think?

I like this idea. I think I can do something along those lines.

packages/data-models/src/definition.js
kirlat: > Then if we decide to use this as a code standard for all out team - let's define - what class properties should be defined as private and what should be defined as public.
> What are exactly conditions to define a class property as private? I suggest to define them and print somewhere :)

On my opinion, it's simple: if the prop is referred to (read or written) outside of the methods of the class, it should be public. If the prop is used by methods of the class only, it should be private. It is very useful during refactoring, for example: we know that private props are used within the class file only, so if we decide to rename it, it would be enough to rename all references within the class. We should be more careful, though, with public members. A public declaration is an indication that this prop is exposed to the outside world and can be used in many places, even outside our own code. So we should change them rarely and be extremely careful while doing so.

Major browsers already start to support the "real" private fields and methods: https://caniuse.com/?search=private. Safari lags behind, unfortunately.

@balmas, @irina060981, what do you think? Does it make sense to you?

kirlat:
@irina060981, it seems I still cannot see your comment within the PR discussion thread, so I'll address it in a separate comment.

> 
```
@@ -44,7 +44,12 @@ export default {
       if (this.$store.getters['app/shortDefDataReady'] && this.lexeme && this.lexeme.meaning && this.lexeme.meaning.shortDefs && this.lexeme.meaning.shortDefs.length > 0) {
         definitionsLocal = this.lexeme.meaning.shortDefs
       } else if (this.lexeme && this.lexeme.lemma.features && Object.entries(this.lexeme.lemma.features).length > 0) {
-        definitionsLocal = [new Definition(this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), 'en-US', 'text/plain', this.lexeme.lemma.word)]
+        definitionsLocal = [
+          new Definition(
+            this.l10n.getMsg('TEXT_NOTICE_NO_DEFS_FOUND'), Language.ENGLISH,
```
> @kirlat , some comments were erased somehow - I would repeat with this ine - because it is a error.
You pass Language.English instead of 'en-US' String, I believe getMsg method does not support passing Object.
You should fix this, I think.

I believe `Language.ENGLISH` is passed to the `Definition` constructor, not to the `getMsg()`, and this is a valid parameter type. But please correct me if I'm wrong here.

packages/data-models/src/language.js
balmas: ok

packages/components/src/vue/components/panel-compact.vue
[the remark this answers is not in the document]
balmas: This is definitely a confusing area, because there are 2 different language values that are pertinent to a definition:
1) the language of the lemma which the definition is a definition of
2) the language of the text of the definition

Currently, our code has the first (the language of the lemma) as a property of the DefinitionSet that contains the Definition and the second (the language of the text of the definition) as a property of the Definition itself.

The language of the text of the definition is set by whatever source that the definition comes from, and in theory could be either an ISO-2 or ISO-3 code.  It may or may not be a language that we have a Language Model class for (i.e. one that Alpheios supports as a source text.)

The question here is related to others raised in this issue - whether the Language object should be used to encapsulate all concepts of Language or just those that we have a Language model for.  I'm not sure of the answer.

packages/data-models/src/definition.js
balmas: I agree we should have some guidelines about how and when to use public vs private properties as well as public vs private methods.

Generally, I think we should avoid public properties where possible, instead using getters and setters.

irina060981: >On my opinion, it's simple: if the prop is referred to (read or written) outside of the methods of the class, it should be public. If the prop is used by methods of the class only, it should be private. It is very useful during refactoring, for example: we know that private props are used within the class file only, so if we decide to rename it, it would be enough to rename all references within the class. We should be more careful, though, with public members. A public declaration is an indication that this prop is exposed to the outside world and can be used in many places, even outside our own code. So we should change them rarely and be extremely careful while doing so.

>Generally, I think we should avoid public properties where possible, instead using getters and setters.

I think that at first we should align our understanding what is public/protected and private in JS context.
According to this - https://javascript.info/private-protected-properties-methods

**public properties/methods** 
- are freely accessable, could be read and updated directly by its name outside the class, 
- they could have have getter/setter, but it is not mandatory
- easily used by child classes via inheritance

**protected properties/methods**
- defined with _ (underscore)
- should not be accessed directly by outside
- getter/setter are mandotary
- could make them readonly
- could be inherited by child classes

**private properties/methods**
- defined with # (hash sign)
- could be accessed only inside the class it is defined
- it could not be inherited in child classes, should use getter/setter inside inheritance

According to the article **protected** is only a programmer convension for now, and **private** is a recent addition to the language. We should have a polyfill for controlling it.
for example like this
https://babeljs.io/docs/en/babel-plugin-proposal-private-methods


So according to all this information and your (@kirlat  and @balmas) proposals I believe we shopuld define the following:

- properties could not be `public` at all
- methods could be `public`

- properties could be `protected` (`_`), should have getter/setter, could be used for inheritance, could be used through methods outside the class
- `protected` properties could be readonly (only getter)
- properties could be `private` (`#`) - should have getter/setter only for inherited classes, and no outside access

- methods could be `protected` - to be used inside the class and in children classes (not outside)
- methods could be `private` - to be used only inside the class (not outside)

- we could define class constants using getters

- we should some polyfil for supporting private

I suggest to define some plan/steps of refactoring the code to this standard.
@balmas and @kirlat , what do you think?

irina060981:
>I believe Language.ENGLISH is passed to the Definition constructor, not to the getMsg(), and this is a valid parameter type. But please correct me if I'm wrong here.

Oh yes, I am wrong here - I am sorry for an  incorrect note!

packages/components/src/vue/components/panel-compact.vue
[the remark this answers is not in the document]
irina060981: According to the current case, I believe, we have several different issues:

We have a language of the targetWord and we have a language of the Definition (_1 issue_) (that should be defined by definition's source and user's locale, but in our case we always have there the same `en-US`)
the language of definition would be used once inside definition vue component (example from short definition):
```
<span ... v-html="definition.text" :lang="definition.language"></span>
```

This Definitions's language is defined in three places:
1. inside lexicons adapter - from the local config file (each lexicon source has a predefined language of the definitions)
2. inside tufts adapter for some languages (lat, syr) - I believe from the remote source
3. inside `panel-compact.vue` (I think it is not a really good place for that) (_2 issue_) - to define an empty Definition (_3 issue_)

I like the following approach:

--Issue 1--
- use Language class as a class for all places where we use Language (for consistency)
- define several getters:
   - getter `code` (instead of toCode() - for better consistency) - it would return default getter `codeISO2` 
   - getter `codeISO2` - `Language.English => 'eng'`
   - getter `codeISO1US` - `Language.English => 'en-US'`
   - getter `codeISO1UK` - `Language.English => 'en-UK'`
- use Language class for Lemma, Definition, Translation, Word Usage and so on, where language inside each of the instance defines the language of saved text in it 

--Issue 2,3-- 
- define defaultLanguage, emptyText, defaultFormat
- define a static method inside Definition object - `createEmptyDefinitionObject(lemmaText)`
- use this static method inside panel-compact instead of direct definition 

Also I think it would be good to define more clearly the defnition's language inside Tufts adapter similiar to Lexicon's adapter

@balmas and @kirlat , what do you think?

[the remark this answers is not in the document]
kirlat: > The question here is related to others raised in this issue - whether the Language object should be used to encapsulate all concepts of Language or just those that we have a Language model for. I'm not sure of the answer.

I tend to think that the `Language` class should be an instance representing just the language code. We probably might not need to force its conversion (but maybe have an option to do that) to any particular ISO 639 format, but rather keep it as provided, as long as it conforms to either of ISO 639‑1, ISO 639‑2, and ISO 639‑3. But, as mappings between those formats are often possible (as https://iso639-3.sil.org/about/relationships states), we can utilize it during comparison, at least. We can also use specific getters to convert from one format to the other, as @irina060981 suggests. Probably it would make sense to use pre-existing libraries for that, such as https://github.com/wooorm/iso-639-3. I don't know if this one is the best, I did not have chance to research it in depth, but I think solutions must exist.

I'm thinking of the language model, on the other hand, as of something bigger than just a language. To me it's more like a set of language concepts that we define for languages we use. So, I think, we use `Language` when we need to store what is the language. We use `LanguageModel` when we need to refer to the extended linguistic information available for the language. That would not always be the case. If we have a definition in German, for example, we'll never need to use the language model for that Same is probably the case for other languages of definition texts: we don't expect to do any transformations that would require extended language data. But if we're working with the lexeme in Latin, for example, we'll need to pull in the Latin Language Model, because without information contained in that object we won't be able to do all the transformations of the lexeme that we need to do. Does it make sense?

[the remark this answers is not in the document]
kirlat: > use Language class as a class for all places where we use Language (for consistency) define several getters:
> * getter code (instead of toCode() - for better consistency) - it would return default getter codeISO2
> * getter codeISO2 - Language.English => 'eng'
> * getter codeISO1US - Language.English => 'en-US'
> * getter codeISO1UK - Language.English => 'en-UK'
> use Language class for Lemma, Definition, Translation, Word Usage and so on, where language inside each of the instance defines the language of saved text in it

I generally agree with this approach, I'm just not sure what getters should return exactly, and how should they be named, consequently. We might not need them at all. A comparison function that can compare two languages with different language codes might be enough. But we should analyze use cases carefully to get the right answer. I think we should try to minimize the amount of code we right and support in order to keep things maintainable.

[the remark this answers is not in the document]
kirlat: > * define defaultLanguage, emptyText, defaultFormat
> * define a static method inside Definition object - createEmptyDefinitionObject(lemmaText)
> * use this static method inside panel-compact instead of direct definition

I might be not fully familiar with all the use cases, but I'm against creating the `Definition` object without a language. If we think the language is a part of the minimal viable object state, and because it's used to calculate the content-based ID, we should always create a `Definition` object with the language defined and never change it afterwards. If it changes, then we need to create a new `Definition` instance instead. That is if we consider `Definition` to be a true value object, which I think it is.

In a situation of a compact panel use case, where we create a definition out of a localized message, we should ideally have a method of `L10n` that would return a language of the currently selected locale, and then try to create a message on that language. If it is not available, we should use the message on the default language. What do you think?

packages/data-models/src/definition.js
kirlat: That's very interesting (thanks for gathering the info!), and I need to think about that a little. I can only say so far that, on my opinion, it's better to avoid Babel for our build process. It will make our build process even more complicated as it is now and I think we should strive to make it simpler instead. It will also increase the build time. It will introduce several dependencies as well. I'm not sure that the benefits it will bring will outweigh that.

balmas: @kirlat and @irina060981 I think what we decided in our check-in today was the following:

Our general development guidelines should be:

* all properties should be "protected/private", accessed by getter/setter methods
* Until all browsers support # private properties we will not distinguish between protected and private in our code
* We will use _ to name private/protected methods and properties
* When full browser support for # private properties/methods is in place, we should reexamine our code to decide which protected/private methods and properties should become explicitly private

Did I miss anything?

kirlat: I think that's all. I will add that to the documentation.

kirlat: > all properties should be "protected/private", accessed by getter/setter methods

I would maybe make an exception from this rule for the pure JSON-like object that contain data only. They often are read only. We might use these objects to represent data passed or received via the GraphQL API, or to store a set of constants. 

Those objects have no methods (thus no logic of their own), and are used for data storage only. I feel that creating a getter and/or setter for each prop of those might be an overkill, especially for objects in GraphQL that could have a lot of fields. What do you think about that? We can use a special naming convention for such object to distinguish them from the full-fledged OOP class instances.

balmas: yes that seems sensible to me

irina060981: agree

Issue #590, opened 2020-12-24
tokenizer adapter - catch and report errors


balmas:
with the fix to alpheios-project/tokenizer#26, the tokenizer service will report client errors (400) in a JSON error object that has the status code and message.  

the tokenizer adapter should catch these and return them to its clients so that they can be displayed to the user. These are likely to be errors with the content the user has supplied totokenize.

500 errors should be caught and reported to its clients as unexpected system errors.

irina060981:
Merged and closed

irina060981:
it is the same as https://github.com/alpheios-project/alignment-editor-new/issues/101
it doesn't need aditional verification - would be verified with https://github.com/alpheios-project/alignment-editor-new/issues/101

irina060981:
Verified with https://github.com/alpheios-project/alignment-editor-new/issues/101

PR #594, opened 2020-12-29
Add message from failed request

irina060981:
For the 
https://github.com/alpheios-project/alpheios-core/issues/590
https://github.com/alpheios-project/alignment-editor-new/issues/101
https://github.com/alpheios-project/tokenizer/issues/26


balmas (approved):
ok, but see request on alpheios-project/alignment-editor-new#106 -- I don't think if requires further changes here since the response code is passed.

irina060981:
>ok, but see request on alpheios-project/alignment-editor-new#106 -- I don't think if requires further changes here since the response code is passed.

ok, I see - will work on it
And will merge this PR

Issue #599, opened 2020-12-29
lemma alignment problems with disamibugation


balmas:
Reported by @rgorman

> I noticed that when I use Alpheios linked to trees, the system gets confused when the lemma is disambiguated by a number, for example, dico2.  It will say that it can't find a definition.
But dico2 is the way the word is offered to the annotator by Arethusa.
Perhaps there is a way to make the two systems more friendly to each other?"

This needs to be done better but it's tricky, because sometimes we DO want to use the number to disambiguate the lemma. But it depends upon the source.  For Greek, we use the same parser (Morpheus) in Alpheios as in the Perseids/Arethusa environment, so we should normally be able to match the enumerated lemma with one of our results.  But in Latin, because Alpheios uses Whitaker while Perseids/Arethusa uses Morpheus, the lemmas are not quite aligned.

Here, retaining the dico2 from the treebank should mean that we actually get the right FULL definition (since we can match dico2 to the dico2 entry in Lewis and Short). So what we probably want is the best of both worlds -- to keep the short definition from Whitaker for dico and and link to the dico2 entry for the full definition.

balmas:
additional use cases from @rgorman

> In the first, the "disambiguated" tag is added to two forms. I hope students don't try to make sense out of "poplar tree," but someone inevitably will.

![image](https://user-images.githubusercontent.com/644870/103412777-cd59b600-4b44-11eb-950d-d34884f2fec3.png)

>  In the second, the numeral on the lemma has led to a "no definition found," and I don't know how I can disambiguate between the two definitions that are offered.

![image](https://user-images.githubusercontent.com/644870/103412826-06922600-4b45-11eb-9a46-7e47a494a887.png)


> In the next example, we see the same phenomenon, and I would really like to be able to tell students that carmen is not a card for wool.

![image](https://user-images.githubusercontent.com/644870/103412856-2f1a2000-4b45-11eb-8eda-5ea76f722c16.png)

> Finally, I would prefer to point my students at the adjective definition for doctus, but the choice is not offered in Arethusa.

![image](https://user-images.githubusercontent.com/644870/103412860-36412e00-4b45-11eb-96a6-6104ce1220b8.png)

For populus, this identifies a gap in our disambiguation logic -- here we could use the masculine gender to disambiguaate between the two lemmas but we instead are creating a masculine inflection the second populus. I think we can add the logic for this. We would still have a problem if there are instances of nouns with the same lemma and gender, but  hopefully that is rare. 

[...]

Issue #540, opened 2020-10-14
treebank: option to show all inflections, highlighting the one from the tree


balmas:
What Alpheios does is disambiguate the lexemes identified by the parser with the lexeme in the treebank.

For the lexemes themselves (i.e. the distinct lemma entries, as far as we can determine them) it identifies which lexeme is the one chosen by the treebank with the little red triangle, but still shows the other lexemes.
 
However, for the inflections within the disambiguated lexeme, it filters the inflections to show *only* the one that was chosen by the treebank.

@vgorman1 requests that we have the possibility to show the various possible inflections and mark the one in the tree with a similar triangle.

balmas:
this is fixed in Alpheios Components 3.3.1-qa.20210106574

For now, the default has been changed to always display all inflections, and to highlight the selected inflection with the same icons we use for the lemma:

![c6f057a5-bbc9-4357-8cde-7030030eb139](https://user-images.githubusercontent.com/644870/103824007-1d52e400-5041-11eb-8952-7acd4c8ed0b4.png)

It's not perfect -- when there are more than 1 possible inflection identified by morpheus, those are displayed below the selected one from the tree, including (again) the one from the tree.  Ideally, I would like to dedupe the inflection from the tree out of the list of the inflections identified by morpheus, but that is too big of a change for me to make right now.

Still to be determined is whether we want this changed default behavior to apply to the treebanked texts at texts.alpheios.net -- need to verify that with @abrasax - but it's ready to test otherwise.

You can use treebanked texts at texts-test.alpheios.net to test as well as the treebank-test page at https://alpheios-misc-dev.s3.us-east-2.amazonaws.com/treebank-test-page/test.html

monzug:
I really like this enhancement!!!

monzug:
in the case of nullis (first sentence of https://texts-test.alpheios.net/text/urn:cts:latinLit:phi0620.phi001.alpheios-text-lat1/passage/1.1), we are adding again the inflection. see screenshot
as Bridget said: It's not perfect -- when there are more than 1 possible inflection identified by morpheus, those are displayed below the selected one from the tree, including (again) the one from the tree. Ideally, I would like to dedupe the inflection from the tree out of the list of the inflections identified by morpheus, but that is too big of a change for me to make right now.

in texts-test version vs the live one
![nullis-cynthia](https://user-images.githubusercontent.com/41396793/104196764-2a315800-53f2-11eb-8b99-01581203b853.JPG)

[...]

PR #604, opened 2021-01-06
I103 i599

balmas:
This PR includes fixes for

#540 
#599 

It makes makes some important changes to the logic for disambiguating lexemes.  We previously were assuming if it was a full homonym match that we should consider the lexeme disambiguated, even if no inflection matched. However there are cases where we might have more than one full homonym match, in which case we can use the inflection to try to get the closest match.

It seems we cannot actually disambiguate a Lexeme outside of the context of the other Lexemes in the Homonym.  

We also have a request (#540) to retain all the inflections in the disambiguated Lexeme, and highlight the one that was disambiguated.

So I have changed the Lexeme.disambiguate method to be one called Lexeme.disambiguateInflections, and this does not change the inflections in the disambiguated Lexeme, nor update it's disambiguation status, but it does set a new selectedInflection property on the disambiguated Lexeme, which contains the disambiguated inflection.

And there is  new method, Lexeme.setDisambiguation which can be used to set the disambiguation flag. The logic of whether to call that is in the Homonym.disambiguate method, because it has to be done in the context of the Homonym.


packages/components/src/vue/components/morph-parts/inflections-list.vue
```diff
@@ -138,15 +170,29 @@
...
+            this.$store.state.app.morphDataReady && this.app.hasMorphData() && this.lexeme.getSelectedInflectionForDisplay)
+            ? this.lexeme.getSelectedInflectionForDisplay()
+            : []
+
+        } else {
+          return (
+            this.$store.state.app.morphDataReady && this.app.hasMorphData() && this.lexeme.getGroupedInflections)
+            ? this.lexeme.getGroupedInflections()
+            : []
+        }
```
balmas, line 195: eventually I think we should dedupe the selected inflection out of the grouped inflections, but this is too big a change for an incremental release.

For annotations, we're going to need to clean this up further, so that inflections can be annotated on distinct from the grouped view of them.

packages/components/src/vue/components/morph-parts/inflections-list.vue
```diff
@@ -230,4 +276,37 @@
...
+      &-icn {
+          fill: var(--alpheios-color-neutral-dark);
+          display: block;
+          height: 22px;
+      }
+  }
+
+  .alpheios-inflections-list__dupe-infl-set {
+      display: none;
+  }
```
balmas, line 311: it may be a site preference whether or not to show the duplicate inflection set. I think the default is not to, but I would like embedding code to be able to override this. Probably we need to expand the treebank data properties to support specification of display preferences, but that was more than I want to do in an incremental release.

packages/components/src/vue/components/morph.vue
```diff
@@ -59,13 +62,24 @@ export default {
...
+    inflectionListClass (lex) {
+      if (lex.getSelectedInflection()) {
+        if (lex.inflections.length === 1) {
+          return "dupe-infl-set"
+        } else {
+          return "super-infl-set"
+        }
+      } else {
+        return "full-infl-set"
+      }
```
balmas, line 74: I am setting these classes here, rather than in the inflection-list component, because its' the duplication of the inflection-list component specified in the morph component that makes it necessary to distinguish between the two with classes. 

I am not sure if we will hide or show the full inflection set in the Alpheios Reader pages, so I want to make this granularly manageable via the css classes (eventually to be moved to data attributes)

packages/data-models/src/homonym.js
```diff
@@ -145,32 +145,59 @@ class Homonym {
...
+          for (const infl of otherLexeme.inflections) {
+            otherLexeme.setSelectedInflection(infl)
+          }
+          missedLexemes.push(otherLexeme)
+        }
       }
     }
     // create a new homonym with the disamibugated lexemes
-    const newHom = new Homonym([...lexemes, ...missedLexemes], base.targetWord)
+    const newHom = new Homonym([...missedLexemes, ...matchedLexemes, ...possibleLexemes, ...unmatchedLexemes], base.targetWord)
```
balmas, line 200: this intentionally changes the order so that the disambiguated lexemes appear first.

packages/data-models/src/language_model.js
```diff
@@ -378,7 +378,7 @@ class LanguageModel {
    * @returns {string} A normalized word.
    */
   static normalizeTrailingDigit (word) {
-    return /^.+1$/.test(word) ? word.substring(0, word.length - 1) : word
+    return /^.+\d$/.test(word) ? word.substring(0, word.length - 1) : word
```
balmas, line 381: this is the fix for #599 (at least partially, there may be more to do for that issue). We should ignore all trailing digits when comparing words, not just a 1

packages/components/src/vue/components/morph-parts/inflections-list.vue
```diff
@@ -138,15 +170,29 @@
       }
     },
     computed: {
+      listClasses () {
+        const klasses = ["alpheios-inflections-list__inflections"]
+        klasses.push(`alpheios-inflections-list__${this.listclass}`)
+        return klasses.join(" ")
```
kirlat, line 176: I'm wondering if the name `klasses` is intentional. Or is it a typo?

packages/data-models/src/lexeme.js
```diff
@@ -50,6 +50,40 @@ class Lexeme {
...
+  /**
+   * Gets the selected inflection formatted for display
+   * (returns an array because the display is grouped by feature
+   * but there should only be one inflection in the array)
+   * @returns {Array} if no selected inflection the array will be empty
+   */
+  getSelectedInflectionForDisplay () {
+    if (this.selectedInflection) {
+      const lm = LMF.getLanguageModel(this.lemma.languageID)
+      return lm.groupInflectionsForDisplay([this.selectedInflection])
```
kirlat, line 83: This method is deprecated, I think it's better to use the static variant of it. We won't also need to get the language model for it.

packages/data-models/src/lexeme.js
```diff
@@ -50,6 +50,40 @@ class Lexeme {
...
+    return this.selectedInflection
+  }
+
+  /**
+   * Gets the selected inflection formatted for display
+   * (returns an array because the display is grouped by feature
+   * but there should only be one inflection in the array)
+   * @returns {Array} if no selected inflection the array will be empty
+   */
+  getSelectedInflectionForDisplay () {
```
kirlat, line 80: I'm wondering if it's better to remove `forDisplay` from the name because it refers to the UI and we want data models to be UI-unaware. Maybe to rename it to something that would reflect how inflections be grouped?

kirlat (commented):
Looks good, but I have a few minor suggestions.

packages/components/src/vue/components/morph-parts/inflections-list.vue
balmas: ah it was just habit to avoid using the word class which is reserved in some languages.

packages/data-models/src/lexeme.js
balmas: ah, okay thanks for that catch!

packages/data-models/src/lexeme.js
balmas: ok, yes you're probably right here. I hate the whole grouping code and would like to rip it out and do it better, which is probably why I avoided the name here. But it is what it is.

packages/components/src/vue/components/morph-parts/inflections-list.vue
balmas: will rename this to classNames

packages/data-models/src/lexeme.js
balmas: I think I still need the language model though, because the method itself is language model class specific (e.g. syriac language model implements this differently than the base language model class)

kirlat: Then, I guess, `lm.constructor.groupInflectionsForDisplay()` should do

balmas: it doesn't work for some reason. Possibly because we don't do the inheritance right. As I really hope this method will go away altogether and be replaced by something better with the lexical query refactoring I'm going to leave it using the deprecated method for now and make sure we have an issue to clean this all up.

PR #607, opened 2021-01-07
fix #589

### LINKED ISSUE #589: mouseMoveEnabled override prevents triggerPreCallback
The logic of `isGetSelectedTextEnabled` in the app controller is wrong.

It currently has:

```
  isGetSelectedTextEnabled (domEvent) {
    return (this.state.isActive() &&
      this.state.uiIsActive() &&
      (!this._options.triggerPreCallback || this.isMousemoveEnabled || this._options.triggerPreCallback(domEvent)))
  }
```

The insertion of the test on this.isMousemoveEnabled before the test on the results of the triggerPreCallback function cause this function to be ignored if isMouseMoveEnabled is true.

balmas:
check on isMouseMoveEnabled should not be in isGetSelectedTextEnabled


packages/components/dist/alpheios-components.js
```diff
@@ -50622,7 +50622,7 @@ class AppController {
   isGetSelectedTextEnabled (domEvent) {
     return (this.state.isActive() &&
       this.state.uiIsActive() &&
-      (!this._options.triggerPreCallback || this.isMousemoveEnabled || this._options.triggerPreCallback(domEvent)))
+      (!this._options.triggerPreCallback || this._options.triggerPreCallback(domEvent)))
```
balmas, line 50625: @irina060981 and @kirlat I cannot see any reason why this test on whether isMouseMoveEnabled should be in this method where we check to see if we can proceed with the getSelectedText event.  As noted in #589 it prevents the triggerPreCallback from working properly if the enableMouseMoveOverride is set on a page using the embedded library. Prior to ee62ccef65c0544af50db0a205d4ad40688095c3  there was a method call here which actually enabled or disabled the mouseMove event, but that was refactored here, and I think in the process there is no reason for the test to be here (and probably it should not have been there before in its earlier form either)  I have tested and if I remove this, the enableMouseMoveOverride in the embedded library still works and now works if triggerPreCallback is set too.

kirlat: It's hard to recall the exact reason to me, but I think the logic here was probably that `triggerPreCallback()` should control the handling of the click events only, but not the `MouseMove`, so that the `MouseMove` be in its own "parallel dimension". But if  `triggerPreCallback()` should affect the `MosueMove` as well then the current logic is incorrect. We don't need the `this.isMousemoveEnabled` check here and it can be removed safely.

I've tested the code with `this.isMousemoveEnabled` removed and it seems to be working well for me.

kirlat (approved):
Looks good to me! Really appreciate the simplification of this convoluted logic

Issue #610, opened 2021-01-12
inflection disambiguation does match (f) in (m,f,n)


balmas:
Wrt to https://github.com/alpheios-project/alpheios-core/issues/540#issuecomment-757999170 there are actually two issues here, one which I didn't see at first

(1)
Morphology Service says the lemma has inflection A and inflection B
Treebank says it is inflection B
We should show inflection B only once in the popup.

Originally I thought that was what was going on with nullis. That has to wait for the changes we're making to support annotations.  However, looking more closely at the output, I realize that is actually a different scenario:

(2) 
Morphology service says inflection A m,f,n
Treebank service says inflection A f
We should recognize that Inflection A f is not different than Inflection A m, f n

This scenario is more similar to #608 and #609.   Will enter a new issue for it.

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/issues/540#issuecomment-758679788_

balmas:
Fixed in Alpheios Components 3.3.3-qa.20210113579

It's as fixed as it can be for now anyway. We won't add a new inflection in this scenario, but you'll still see the inflection from the tree along with the with from the morphology service.

That is, for example, if the morphology service parse says the gender of a form is all genders (i.e. m, f, n) and the treebank says it's feminine, the treebank inflection will be pulled out as the disambiguated one but it won't result in a new inflection being added to the lexeme.

So you'll get 
![ad482416-843f-411e-a9a3-a823a27d4f61](https://user-images.githubusercontent.com/644870/104509886-5696d100-55b8-11eb-891a-c47abe816ea9.png)

But not 
![prior](https://user-images.githubusercontent.com/41396793/104196764-2a315800-53f2-11eb-8b99-01581203b853.JPG)

monzug:
Super. tested in  Alpheios Components 3.3.3-qa.20210113579

Issue #609, opened 2021-01-11
Greek particle == Greek adverb


monzug:
go to line 46 of https://texts-test.alpheios.net/text/urn:cts:greekLit:tlg0012.tlg001.alpheios-text-grc1/passage/1.1-1.30,
both words δ᾽   and ἄρ᾽ are in the tree but we do not show any treebank icons in the pop-up. this is also on texts.alpheios.net

balmas:
this is similar to #608 

upon disambiguation of inflections we need to be able to match a greek particle to a greek adverb

balmas:
Alpheios Components 3.3.3-qa.20210113579

The code fix was to all a particle (i.e. as reported by morpheus) to be disambiguated by an adverb from a tree.

E.g. now we will get

![d2010edb-6ec6-480e-9b18-1b22e5d1c7a4](https://user-images.githubusercontent.com/644870/104510419-269bfd80-55b9-11eb-8c52-124ac2992a16.png)

for the above example

monzug:
cool. to clarify, it's not that all adverb are now particle. adverb are still present e.g. αὐτοῦ   in the same tree as the example in this issue. this fix is to match a greek particle to a greek adverb, if a greek word has both in the tree (e.g. δ᾽ ἄρ᾽)

balmas:
@monzug please reverify in Alpheios Components 3.3.3-qa.20210115626 (changes for #615 directly impacted this fix)

monzug:
question: have interjection and exclamation also merged? see latest build vs release

![Screen Shot 2021-01-18 at 4 33 06 PM](https://user-images.githubusercontent.com/41396793/104934905-e1a60b80-59aa-11eb-86aa-9d67aeac0f30.png)

![Screen Shot 2021-01-18 at 4 32 29 PM](https://user-images.githubusercontent.com/41396793/104934910-e2d73880-59aa-11eb-8ea2-63637bee591e.png)

balmas:
oh, this is a good catch! No, that's actually a bug.  I need to fix the bug.

But you're right, it's better when we consider exclamation and interjection the same. Do you think there is any reason not to?  Is  exclamation a more granular form of an interjection ?  (@vgorman1 and @rgorman I welcome your input on this question - Whitaker parses agedum as an interjection; it's treed as an exclamation. Is there ever a reason to say an exclamation is not an interjection?)

vgorman1:
All interjections is best.

[...]

Issue #608, opened 2021-01-11
Latin verb with mood=gerundive matches whitaker mood=participle


balmas:
another interesting point here, which @rgorman helped clarify -- the treebank reports this as mood=gerundive. The Whitake parser used by Alpheios reports all Gerunds as verb participles (see alpheios-project/morphsvc#11).  

I think perhaps we should make a change to the Alpheios treebank adapter to consider a Latin verb with mood=gerundive as being the same as a verb participle. That way we will at least be comparing apples to apples. We'd still have a disconnect here, because the tense in the treebank is present. But we would at least be matching the part of speech.

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/issues/540#issuecomment-758232342_

balmas:
this is fixed in Alpheios Components 3.3.3-qa.20210113579

the code change was to consider a gerundive the same as a verb participle -- @monzug note that for now this change applies only to the treebank disambiguation and not the inflection matching.

There is also a corresponding change to the treebank data alpheios-project/alpheios-trees#15  to remove the incorrect tense

monzug:
fugiendo (https://texts-test.alpheios.net/text/urn:cts:latinLit:phi0620.phi001.alpheios-text-lat1/passage/1.1) is only verb participle fut. pass. - the present form is gone

![Screen Shot 2021-01-14 at 1 08 59 PM](https://user-images.githubusercontent.com/41396793/104589446-2ad42380-566a-11eb-9b79-2de7adc4ac4e.png)

balmas:
I went a little too far with this change, per #615 

It's not actually that Whitaker reports the whole lexeme as being a verb participle when an inflection has the gerundive mood, but rather that the inflection itself has a part of speech which is verb participle but the lexeme itself is a verb.  So I need to make this change a little less encompassing and more specific:
1) a gerundive mood to match a participle mood.  
2) when retrieving the data from the treebank, set the pofs of the inflection of a gerundive to verb participle instead of verb

balmas:
@monzug please reverify in Alpheios Components 3.3.3-qa.20210115626 (changes for #615 directly impacted this fix)

monzug:
retested in Alpheios Embedded Library 3.3.1-20200821383
Alpheios Components 3.3.3-qa.20210119683  on Mac

PR #613, opened 2021-01-13
I608 609 610 infl disambig

balmas:
this PR introduces fixes for #608 #609 and #610 , all problems with disamibugation of lexemes  and inflections produced by the morphological parses with treebank data.

Some of this may duplicate logic that already exists in our inflection-tables library for matching between parser inflections and the inflection-table data, but it did not make sense to try to reuse that logic here because as we move forward, we're going to have to handle an increasing number of disparate sources of lexical data.  The business logic for normalization and matching is going to need to reside in the domain along with the language models.  Eventually we will have to refactor the inflection-tables code as well.

So these changes certainly don't cover every mismatch scenario we might come across with treebanked data, but they are a start and are extensible. They also don't affect any other part of the application for now.

We will need to be thinking about how this all fits into the the lexical query refactoring.

I have not tThere is already some pretty complex logic in the inflection-tables library for dealing with mat


packages/data-models/src/inflection.js
```diff
@@ -214,28 +214,61 @@ class Inflection {
...
+   *                   an exactMatch means the disamibugator matched all
+   *                   values of all features
    */
-  disambiguatedBy (infl) {
+  disambiguatedBy (infl, ignorePofs = false) {
     let matched = true
+    let exactMatch = true
     // an inflection can only be disambiguated by its features
-    if (this.features.length === 0 || infl.features.length === 0) {
+    if (this.features.size === 0 || infl.features.size === 0) {
```
balmas, line 246: this test actually wasn't working before, because `features` is a Set and doesn't have a `length` property.

packages/data-models/src/inflection.js
```diff
@@ -214,28 +214,61 @@ class Inflection {
...
    * @param {Inflection} infl Inflection object to be used for disambiguation
+   * @param {Boolean} ignorePofs flag to ignore the inflection's part of speech
+   *                             (use if lexeme pofs is more relevant)
+   * @returns {Object} object { {Boolean} match, {Boolean} exactMatch }
+   *                   a match means the inflection was disamibugated
+   *                   an exactMatch means the disamibugator matched all
+   *                   values of all features
    */
-  disambiguatedBy (infl) {
+  disambiguatedBy (infl, ignorePofs = false) {
```
kirlat, line 242: I would suggest to put disambiguation options into an options object so that we can add other options without adding extra parameters and changing the method's signature: `disambiguatedBy (infl, { ignorePofs = false } = {})`. What do you think?

packages/data-models/src/inflection.js
kirlat: An excellent catch!

packages/data-models/src/inflection.js
```diff
@@ -316,9 +349,9 @@ class Inflection {
    * @param {string} featureValue - A value of a feature
    * @returns {boolean} True if an inflection contains a feature, false otherwise
    */
-  hasFeatureValue (featureName, featureValue) {
+  hasFeatureValue (featureName, featureValue, normalize=false) {
```
kirlat, line 352: Maybe put `normalize` into an options object as well?

packages/data-models/src/language_model.js
```diff
@@ -452,6 +473,22 @@ class LanguageModel {
...
 
+  /**
+   * Compare two feature values with language specific logic
+   *
+   * @param {string} featureType - the feature type being compared
+   * @param {string} valueA - the first value for comparison
+   * @param {string} valueB - the second value for comparison
+   * @param {boolean} normalize - whether or not to apply normalization
+   */
+  static compareFeatureValue ( featureType, valueA, valueB, normalize = true) {
```
kirlat, line 484: Maybe add `normalize` to the options object here too? We might have other options in the future.

kirlat (approved):
Looks good, I just think it might be better to pack options in the single extensible object.

kirlat:
I'm wondering to what extent those "normalization" rules are universal. Could it happen that the normalization that makes sense for the treebank would produce incorrect results for the other source? In that case maybe we should add a reference to the source to take that into consideration?

balmas:
> Looks good, I just think it might be better to pack options in the single extensible object.

Yes I agree. I'll make those changes.

balmas:
> I'm wondering to what extent those "normalization" rules are universal. Could it happen that the normalization that makes sense for the treebank would produce incorrect results for the other source? In that case maybe we should add a reference to the source to take that into consideration?

This is a complicated topic. I did note in a few places the rationale for the current decisions in source comments. It certainly is conceivable that normalization logic will need to be more complex and  dependent upon the source. It's hard to say without encountering actual data.

Up until now, we've more or less taken the approach that lexical attributes need to be normalized before coming in to Alpheios, and the client adapters are where we've done that (as well as in the inflection-table library which serves as a sort of adapter for the inflection table data).  But as we move towards allowing user created annotations, and additional sources, I think we need to be able to respect the choices made in those sources, and display them as provided in the interface, but still have a base which everything gets normalized to for comparison purposes.

Issue #620, opened 2021-01-19
Ignore a specific character around a Greek word 


monzug:
Double click does not work for  †οὖνπερ
ἱεροδόκα†  in  https://texts.alpheios.net/text/urn:cts:greekLit:tlg0085.tlg001.alpheios-text-grc1/passage/1-30    line 362 and 363 because of the †.  could we ignore it?

balmas:
yes this would be easy but it needs to wait for the next release.

monzug:
No problem. Thanks

balmas:
please make the fix in a branch off of the incr-3.3.x branch

kirlat:
Has been merged into ` incr-3.3.x`.

irina060981:
It should be implemented (it was merged to the code before)

@monzug , could you check?

PR #624, opened 2021-01-22
A punctuation fix

kirlat:
Added a dagger and a double dagger to the list of punctuation characters of all languages. In response to #620.

There is also a bunch of ESLint auto-correction that I decided to leave because they, on my opinion, make sense.


balmas (approved):
thank you!

PR #627, opened 2021-01-26
A rollback of install-peerdeps

kirlat:
A rollback of `install-peerdeps` version to fix alpheios-project/node-build#45. I think simply specifying the version in the npx command should be enough.


balmas (approved):
great, thanks.

Issue #623, opened 2021-01-21
Remove numbers from lemma in wordlist


monzug:
Remove numbers from lemma in wordlist, see #599 for reference.
double click on omnia in https://texts-test.alpheios.net/text/urn:cts:latinLit:phi0959.phi006.alpheios-text-lat1/passage/1.1-1.30
the lemma is omnis1, omne, omnia in wordlist. also when limit to lemma, the number is still in the lemma. (note that the lemma in wordlist from a lookup word does non have any number) 

![Screen Shot 2021-01-21 at 2 42 06 PM](https://user-images.githubusercontent.com/41396793/105363712-a26ef900-5bfc-11eb-8ee6-fad7a924d03f.png)

lemma in wordlist from lookup word
![Screen Shot 2021-01-21 at 3 27 28 PM](https://user-images.githubusercontent.com/41396793/105364248-3e990000-5bfd-11eb-8f90-4107bd9827b1.png)

balmas:
please make the fix in a branch off of the incr-3.3.x branch.

kirlat:
This issue seems to be related to the usage of Treebank data. It does not occur when the Treebank data is not involved.

I am able to reproduce it and is on my way fixing it.

monzug:
Yes, it's treebank related. see also #599

kirlat:
The variant returned by the Treebank is `omnis1`, and due to how our current disambiguation works, we prefer that variant for the `lemma.word` over the Tuft's value of `omnis`.

The disambiguation function that decides this is https://github.com/alpheios-project/alpheios-core/blob/incr-3.3.x/packages/data-models/src/lemma.js#L221-L266. Both words has no mixed cases and both do not need normalization, so we skip any logic related to that. But the `otherLemma` (which is the treebank's value) happen to have a digit at the end, so we prefer it over the Tuft's value: https://github.com/alpheios-project/alpheios-core/blob/incr-3.3.x/packages/data-models/src/lemma.js#L258.

With `lexeme.setDisambiguation()` https://github.com/alpheios-project/alpheios-core/blob/incr-3.3.x/packages/data-models/src/lexeme.js#L238-L243 we replace the `lemma.word` value only, but `lemma.prinicpalParts` field is not update and remains the same as it is in the Tuft's results (i.e. with `omnis`, not `omnis1`).

As a result of the disambiguation process, the first lexeme of the word contains the following information:
```
lemma.word: 'omnis1'
[...]

PR #630, opened 2021-01-28
Removal of tralining digits from lemmas in the word list

kirlat:
Removed trailing digits from lemmas displayed in the word list according to #623.


packages/data-models/src/lemma.js
```diff
@@ -45,7 +45,12 @@ class Lemma {
   }
 
   get displayWord () {
-    return this.word.replace(/\d+$/, '')
+    return Lemma.getDisplayWord(this.word)
+  }
+
+  static getDisplayWord (word) {
+    // In the word list we meed to format a word that is in a string format so the getter of the Lemma won't work there
+    return word.replace(/\d+$/, '')
```
irina060981, line 53: If to be honest I didn't understand the sense of this refactoring.
The method display word could depend not only on the word of the Lemma, but other properties like language.
By its sense it is not a method of abstract lemma - it is a getter of the current lemma with its current word.
Do you suggest to use this method for any Strings to remove digits?

packages/components/src/vue/components/word-list/word-item-panel.vue
```diff
@@ -127,6 +130,9 @@ export default {
     },
     setLemmaFilterByClick(lemma) {
       this.$emit('setLemmaFilterByClick', lemma)
+    },
+    formattedWord (lemma) {
+      return Lemma.getDisplayWord(lemma)
```
irina060981, line 135: And the same here (my question is in the next note) - I don't understand why do we need these additional code.
From my point of biew it is  a commonly used getter for a specific instance of Lemma, but not an abstarct formatter of any String.

kirlat: We're using `displayWord` to get the lemma's word formatted for the display in the popup. We need to apply the same rules to display lemma words in the wordlist (@balmas please correct me if I'm wrong). However, we do not have `Lemma` objects in the wordlist, lemma words are presented as strings there. Because of that, using a Lemma's getter is not an option. So I came up with with a static method that can be used to format a string representing a lemma's word. It's not supposed to be used with any string, but only with those that represent lemma words. I agree it's not ideal but I don't see a better solution now.

I believe it's more than just removing the trailing digits, its to make lemmas in the wordlist be formatted the same way as in other places of the app, so I think we should use the same formatting routine for both.

Do you have any suggestion how to do it differently, considering a limited amount of time that can be spent on this?

packages/data-models/src/lemma.js
kirlat: Please see my other comment: https://github.com/alpheios-project/alpheios-core/pull/630#discussion_r566810975

I agree that it could depend on language too, but in that case I think we can add a `language` parameter to the static method and that would solve it.

Do you have any suggestions how to do it differently? I think it's not just about removing the trailing digits (we could use a local method in the wordlist for that), it about applying the same formatting rules to words of lemmas across the app. So I think those rules should be defined in the central location. Once the rules are changed, the change should be applied to all parts of the applications.

balmas (approved):
@kirlat 's changes look okay to me. His assessment of why this is needed is correct from my point of view.

packages/components/src/vue/components/word-list/word-item-panel.vue
irina060981: In the `word-item-panel.vue` we use 
```
lemmasList () {
  return this.$store.state.app.wordListUpdateTime ? this.worditem.lemmasList.split(',') : []
}
```
in the wordItem class we could see that
```
  get lemmasList () {
    if (this.homonym && this.homonym.lexemes) {
      return this.homonym.lexemes.map(lexeme => lexeme.lemma.word).filter((value, index, self) => {
        return self.indexOf(value) === index
      }).join(', ')
    }
    return ''
  }
 ```
 
 So this way you could see, that I used lemmas but not some simple words.
In this case I suggest to create `displayLemmaList` - and use `lemma.displayWord` instead of `lemma.word`
 
 I think it would be more consistent
 
 Hope it will help

kirlat: I thought of using `Lemma` objects inside the word item Vue component but decided it would be too complex because the changes will have the ripple effect. But if we create a `Lemma` object inside the `formattedWord()` method then all the changes would be isolated there. That would allow to keep it simple and yet avoid adding a static method to the `Lemma`. So that's what I did. Please let me know what do you think.

irina060981: I didn't understand, why you didn't decide to simply add a new display method for lemmasList inside `data-models` WordItem` class.
It could be used in other places in that case.
From my point of view - your decision is more complex.

But if you think that your solution is better , then you could do as you think.

kirlat: > I didn't understand, why you didn't decide to simply add a new display method for lemmasList inside data-models WordItem` class.
It could be used in other places in that case.

I was not sure if it would be useful in other places, but that's maybe because I don't have all the full picture of all the usages of the wordlist. The current fix keep changes rather minimal and I think it might be good this way. We're planing to refactor wordlist later so we can postpone the other changes till then.

irina060981: It is also used inside download of the wordlist
`download-confirmation.vue`

If you apply your solution - then you would fix inside the wordlist panel, but download would be with digits.

That's why - I suggest to create only one getter and used in both places

kirlat: > If you apply your solution - then you would fix inside the wordlist panel, but download would be with digits.

If I understand correctly, we need to remove digits from the display only. We would still want to keep digits in all our data structures internally. @balmas, is that correct?

balmas: yes that's correct.

Issue #612, opened 2021-01-13
more granular highlighting of selections


balmas:
when looking up ἐπεκυλίνδουν and selecting one of the compound word parts in the pop-up, can we highlight only the compound part that has been selected ( ἐπε or κυλίνδουν ) not the all word?

_Originally posted by @balmas in https://github.com/alpheios-project/alpheios-core/issues/541#issuecomment-759549799_

irina060981:
It should be already implemented.
@monzug , could you check?

monzug:
it took me a while to understand what exactly I was supposed to test but I figured it out and it has been fixed.
tested in  Alpheios Reading Tools 3.4.1 build incr-3.4.x.20211125333
Alpheios Components 3.3.3-incr-3.4.x.20211123271 build ext-i323-message.20211125609

irina060981:
Great - it was not very clear for me too :)

PR #631, opened 2021-01-28
Made selection of compound words more granular

kirlat:
I think replacing `span` with `div` would do it. For #612.


balmas (approved):
ok thank you

Issue #628, opened 2021-01-27
Wordlist: number of times looked up a word after login set to 1


monzug:
@irina060981 question for you: why does the number of times you have looked up a word in wordlist go back to 1 after login?

e.g. as guest user, double click on few latin words 2 or 3 times, then log in and go to wordlist and limit to this session.
the number of times you have looked up a word in wordlist goes back to 1 

as guest user
![Screen Shot 2021-01-27 at 2 57 53 PM](https://user-images.githubusercontent.com/41396793/106001735-9da0be00-60b0-11eb-9644-1450fe986a8c.png)

after login (note that caro was already saved in wordlist)
![Screen Shot 2021-01-27 at 3 00 09 PM](https://user-images.githubusercontent.com/41396793/106001856-be691380-60b0-11eb-86af-9bb33dabaf55.png)

monzug:
maybe this would be a question for @kirlat

irina060981:
I work now only with alignment-editor - but I could check - as I was creating this code

monzug:
Irina, Thanks. it's just a question so if it's not too much of a burden for
you, it would be nice. Thanks

irina060981:
@balmas, I found the problem - and created PR for fixing it

monzug:
Great! Thanks

irina060981:
Merged and closed (to both master and incr-3.3.x)

monzug:
@balmas, for future releases

PR #633, opened 2021-02-01
Small fix for frequency update in wordlist

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/628

It is a small fix for frequency update on merging data from local and remote

@balmas and @kirlat - it is a small fix (only one code line) that doesn't influence on other code
I would merge it without a review


balmas:
@irina060981 if this bug is in the production code, can you please merge the fix into the incr-3.3.x branch as well?

Thank you!

irina060981:
I have skipped some tests that uses remote sources (they failed) and fix one (outdated)

Issue #639, opened 2021-03-10
Add ability to define language of the text from the character's set


irina060981:
- investigate libraries/methods to define language for the lookup words from character's set
- add ability to define language from the character's set
- merge it to production `incr-3.3.x`

irina060981:
Priorities for language definition:
1. html markup
2. language detection from unicode
3. default language

if a user updates language in a popup for empty lexical query - it would have the biggest priority

irina060981:
@monzug , this is ready for testing

monzug:
I uploaded a file with Arabic text set as Italian. the language detection failed, offering to choose a new language from drop-down. As Arabic is one of the supported languages in pop-up, I was expecting to be able to detect the language. maybe, I misunderstood this fix. see attachment

![Screen Shot 2021-03-30 at 2 57 55 PM](https://user-images.githubusercontent.com/41396793/112993331-65c5fd80-9169-11eb-8cf0-243b85c4a768.png)

monzug:
still to test in web extension

irina060981:
>question 1): if I have the wrong language in markup (priority 1), priority 2 will never kick in, right? the wrong language has priority 1, correct? the only option to change language is in pop-up or by changing language in Origin or Target texts.

quite right

irina060981:
>Not sure if it would make sense to reverse the priorities as follow: application detects the language (as priority 1), if not able to detect the language then it takes the one from markup (priority 2)? at the meeting we all agreed in the above workflow, right?

We talked about this on the meeting and agreed, that we should consider - if a language was defined in HTML markup, than we should trust it more than language detection

irina060981:
>question 2): how can I test a scenario in which there is no lang in markup?

With the help of webextension I believe

monzug:
> > question 2): how can I test a scenario in which there is no lang in markup?
> 
> With the help of webextension I believe on the sites without html markup.
> I created a specific index.html and test with its help.
> @monzug , I published it for you - may be it would be helpful
> https://alpheios-misc-dev.s3.us-east-2.amazonaws.com/alignment-editor/index-detect.html

Thanks Irina, I did also use the test.html file that is very similar to yours

monzug:
> still to do: test in web extension

test on build  Alpheios Reading Tools 3.3.3 build incr-3.3.x.20210331291
Alpheios Components 3.3.3-incr-3.3.x.20210326.58 build i639-language-detect-2.20210331450  in  FF/PC

[...]

PR #641, opened 2021-03-25
Define language from detection and with priority of default

irina060981:
For the issue https://github.com/alpheios-project/alpheios-core/issues/639

- added language detection (from unicode) - https://github.com/alpheios-project/language-detect
- added prioritizeDefaultLanguage once after it was chosen in popup

**The logic from user perspective:**
A user lookups for a word in a text by double-click.
There is a workflow for defining a language for the word:
- there could be a language that defined in HTML markup (priority 1)
- an application detects language from the text by analyzing unicode (language-detect library) (priority 2)
- there is a defaultLanguage defined in options (priority 3)

For example:
`ἔννεπε`
- there is `grc` in markup
- there is detected `grc` from the detection
- there is `lat` as a default
=> whould be chosen `grc` because of markup

For example:
`ἔννεπε`
- there is no lang in markup
- there is detected `grc` from the detection
- there is `lat` as a default
=> whould be chosen `grc` because of detection

But we could have the following situation:
`مَقَرٍ`  (Arabic)
- there is no lang in markup
- there is detected `per` from the detection (it is a common problem to separate Arabic from Persian)
- there is `lat` as a default
=> whould be chosen `per` because of detection, but it is wrong and there would be no results for lookup.
Popup would suggest to update defaultLanguage. And if a user changes it - it would have the highest priority

`مَقَرٍ`  (Arabic)
- there is no lang in markup
- there is detected `per` from the detection
- there is `ara` as a default (changed by the user)

=> whould be chosen `ara` because of default changed by a user
After this highest priority is cancelled (it is used only for the one selection)

**The logic from code perspective:**
There is an event of textSelection -> SelectionController.onTextSelected executes
It creates `const htmlSelector = new HTMLSelector(event, defaultLangCode)` and saves 
- `defaultLanguage`
- `languageFromSource`
Then it creates `const textSelector = htmlSelector.createTextSelector()` and extracts text of the selection
Then we get the current state of `prioritizeDefaultLanguage` and executes `htmlSelector.defineLanguage(textSelector, prioritizeDefaultLanguage)`
It calculates the final language
Then it publishes `SelectionController.evt.TEXT_SELECTED` and inside this event - it cleares `prioritizeDefaultLanguage` 

If there is no result - then we have a notification inside a popup with a select of defaultLanguage (it is called prefferedLanguage)
If it is changed - then an application commit a change for `prioritizeDefaultLanguage`  - set it as `true`
And if a user does the second lookup it would get `prioritizeDefaultLanguage` as true

Some bug fixes:
- added key for v-for
- remove error for long greek definitions upload


packages/client-adapters/dist/alpheios-client-adapters.node.js
```diff
@@ -5629,7 +5629,7 @@ class AlpheiosLexiconsAdapter extends _clAdapters_adapters_base_adapter__WEBPACK
           lastAlt.push(strippedAll[0])
         }
       }
-      if (lastAlt.length > 0) {
+      if (data && lastAlt.length > 0) {
```
irina060981, line 5632: Added for the cases when data is uploaded from remote source too long, and user has time to request a new query

packages/client-adapters/src/adapters/lexicons/adapter.js
```diff
@@ -485,7 +485,7 @@ class AlpheiosLexiconsAdapter extends BaseAdapter {
           lastAlt.push(strippedAll[0])
         }
       }
-      if (lastAlt.length > 0) {
+      if (data && lastAlt.length > 0) {
```
irina060981, line 488: Added for the cases when data is uploaded from remote source too long, and user has time to request a new query.
It is a bug fix for showing error when data is not ready (it is not for the min task of the PR)

packages/client-adapters/tests/adapters/lexicons/adapter.test.js
```diff
@@ -266,7 +266,7 @@ describe('lexicons/adapter.test.js', () => {
     }
   })
 
-  it('12 AlpheiosLexiconsAdapter - checkCachedData execute fetch, fillMap only once - if successed', async () => {
+  it.skip('12 AlpheiosLexiconsAdapter - checkCachedData execute fetch, fillMap only once - if successed', async () => {
```
irina060981, line 269: It is a test for remote source that failed because of timeout

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -121,7 +121,10 @@ export default class AppController {
      * @type {UIEventController}
      */
     this._evc = new UIEventController()
-    this._selc = new SelectionController(this.getDefaultLangCode.bind(this))
+    this._selc = new SelectionController({
+      getDefaultLangCodeFn: this.getDefaultLangCode.bind(this),
+      getPrioritizeDefaultLanguageFn: this.getPrioritizeDefaultLanguage.bind(this)
+    })
```
irina060981, line 127: This is the controller that operates with text selection.
These methods are closures of Vuex.Store variables.
I passed this way because SelectionController does not have a reference to the store.
And I decided to integrate it the same way it was done before

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -493,7 +496,8 @@ export default class AppController {
         providers: [], // A list of resource providers
         queryStillActive: false, // it is for Persian case, when we canReset
 
-        mouseMoveOverrideUpdate: 1
+        mouseMoveOverrideUpdate: 1,
+        prioritizeDefaultLanguage: false
```
irina060981, line 500: This is a Vuex.Store variable that controlls when default language should ghave the highest priority

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -643,6 +647,10 @@ export default class AppController {
 
         setMouseMoveOverrideUpdate (state) {
           state.mouseMoveOverrideUpdate = state.mouseMoveOverrideUpdate + 1
+        },
+
+        setPrioritizeDefaultLanguage (state, value = true) {
+          state.prioritizeDefaultLanguage = value
```
irina060981, line 653: Vuex setter

packages/components/src/vue/components/inflections/inflections-browser.vue
```diff
@@ -16,13 +16,13 @@
         <wide-table
             :collapsed="inflBrowserTablesCollapsed"
             :standard-form-data="table"
-            v-for = "(table, index) in getTables('lat', 'noun')" 
+            v-for = "(table, index) in getTables('lat', 'noun')" :key="getIndex(index, 'lat', 'noun')"
```
irina060981, line 19: The following changes - bug fixes (not for the main change)

irina060981:
@zfletch , I would merge this branch to prepare a build for Monica's testing.
But feel free to review the PR I would implement your suggestions in the next PR with Monica's feedback.

zfletch (commented):
This looks mostly good to me. I'd recommend rewriting `getLanguageID` as a pure function (see comment).

I do have some concerns about the language detection. I'll leave them below instead of opening an issue in the `language-detect` repo.

---

The code to detect languages is below:

```javascript
export default class LanguageDetect {
  static get langs () {
    return {
      grc: { regex: /^[\u0374-\u03FB ]|[\u1F00-\u1FFE]+$/u, langCode: 'grc' },
      gez: { regex: /^[\u1200-\u137C ]+$/u, langCode: 'gez' },
      zho: { regex: /^[\u4E00-\u4EFF ]|[\u5000-\u5FFF]|[\u6000-\u6FFF]|[\u7000-\u7FFF]|[\u8000-\u8FFF]|[\u9000-\u9FFF]+$/u, langCode: 'zho' },
      syr: { regex: /^[\u0700-\u074F ]|[\u0860-\u086A]+$/u, langCode: 'syr' },
      per: { regex: /^[\u0621-\u06CC ]+$/u, langCode: 'per' },
      ara: { regex: /^[\u0600-\u06FF ]|[\u0750-\u077F]|[\u08A0-\u08FF]|[\uFB50-\uFDFF]+$/u, langCode: 'ara' },
      lat: { regex: /^[A-Za-z ]+$/, langCode: 'lat' }
    }
  }

  static get defaultLangCode () {
    return 'lat'
  }

  static detect (text) {
    let langRes
    for (let i = 0; i < Object.values(this.langs).length; i++) {
      const langData = Object.values(this.langs)[i]

      if (langData.regex.test(text)) {
        langRes = langData.langCode
        break
      }
    }
    return langRes
  }
}
```

1. I think there's a bug in the regular expressions. The `^` operator binds more tightly than `|`, which means that the Greek regex returns true if the first character matches `[\u0374-\u03FB ]` or if any character matches `[\u1F00-\u1FFE]+`. For example, the strings `'Φ not Greek'` and `'not Greek ἀ'` match.

2. The loop is calling `Object.values` in every iteration, which makes it slower (it changes it from O(n) to O(n<sup>2</sup>)). This doesn't make much a difference because the number of items is small, but it can be fixed simply by moving the `Object.values` call outside the loop:

```javascript
const languages = Object.values(this.langs);
for (let i = 0; i < languages.length; i++) {
  const langData = languages[i];

 ...
}
```

3. I'd also recommend either making `langs` an array or removing `langCode`. Unless I'm missing something (always possible), it's redundant because it has the same value as the object key:

```javascript
// If you change langs to an array
return langs.forEach(({ regex, langCode }) => {
  if (regex.test(text)) {
    langRes = langCode;
    return;
  }
})

// If you make use of the key
const keys = Object.keys(this.langs);
for (let i = 0; i < keys.length; i++) {
  const key = keys[i];
  const langData = this.langs[key];

  if (langData.regex.test(text)) {
    return key;
  }
}
```

packages/client-adapters/tests/adapters/lexicons/adapter.test.js
irina060981: I think that it should be used when `AlpheiosLexiconsAdapter` is updated.
But it is too heavy for each PR. That's why I skipped it but not deleted.

packages/components/src/lib/selection/media/media-selector.js
```diff
@@ -28,11 +28,15 @@ export default class MediaSelector {
    * @return {symbol} A language ID of a selection
    */
   getLanguageID (defaultLanguageCode) {
-    const code = this.getLanguageCodeFromSource() || defaultLanguageCode
+    this.defaultLanguageCode = defaultLanguageCode
+    this.languageFromSource = this.getLanguageCodeFromSource()
```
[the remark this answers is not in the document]
irina060981: Agree to your comment.
I checked that this method is not widely used, so - I believe I could update it.
Thank you for the advice.

irina060981:
> This looks mostly good to me. I'd recommend rewriting `getLanguageID` as a pure function (see comment).
> 
> I do have some concerns about the language detection. I'll leave them below instead of opening an issue in the `language-detect` repo.
> ```

@zfletch, thank you for you advices for the language-detection repo!

1. Thank you for catching error with regexps. Didn't check it in tests. Will fix!

3. I have created it as a dictionary to be easy updatable. But after your comment - I agree it is redundant. I will update it to array and this way it will fix 2.

Issue #644, opened 2021-04-12
Add adapter for language detection


irina060981:
We need a new adapter for the API - https://detectlanguage.com/documentation
It would be used for language detection for texts in Alignment Editor

PR #645, opened 2021-04-14
Add adapter for language detection

irina060981:
For the issue #644 

- added adapter for detect language API https://detectlanguage.com/documentation

**There are short explanations of the package's structure.**

`src/adapters/adapters-config.json` - has all adapter's description and properties, this would be checked on it ClientAdapter's initialization
`src/adapters/base-adapter.js` - has all fetch methods for all adapters
`src/client-adapters.js` - has all public methods for `client-adapters`

Each adapter has its own folder with main file `adapter.js`.

Initialization examples could be found in tests.

Most of adapters support passing stub data for excluding usage of real remote requests (to reduce traffic or save allowed amount of data)


irina060981:
Also There are changes for the 
https://github.com/alpheios-project/alpheios-core/issues/639
https://github.com/alpheios-project/alignment-editor-new/issues/242
https://github.com/alpheios-project/alignment-editor-new/issues/302

Fixed to the following behaviour:
- if defined language is unsupported - use default language (as it was before)
- if a user chose another language it is used as priority without limitation for the page

@zfletch , it is a fix for the PR you've reviewed before
https://github.com/alpheios-project/alpheios-core/pull/641

packages/client-adapters/src/adapters/detectlang/adapter.js
```diff
@@ -0,0 +1,84 @@
...
+    return null
+  }
+
+  /**
+   * The remote service returns the following format
+   * { data: {
+        detections: [
+          { language: 'en', isReliable: true, confidence: 3.36 },
+          { language: 'pt', isReliable: false, confidence: 3.36 },
+          { language: 'eu', isReliable: false, confidence: 3.36 }
```
zfletch, line 64: I believe two-letter codes are ISO 639-1 (not 3 or  639-3).

packages/client-adapters/src/adapters/detectlang/adapter.js
```diff
@@ -0,0 +1,84 @@
...
+      }}
+   * We need return only one the most reliable languageCode in ISO-3 format
+   * @param {Object} langsData
+   * @returns {String|null} lang code in ISO-3
+   */
+  chooseOneLanguage (langsData) {
+    if (langsData && langsData.data && langsData.data.detections && langsData.data.detections.length > 0) {
+      const reliableLangs = langsData.data.detections
+        .filter(langItem => langItem.isReliable)
+      if (reliableLangs) {
```
zfletch, line 75: This may not behave as expected. The empty array is `true` in JavaScript.

packages/client-adapters/tests/client-adapters.test.js
```diff
@@ -672,4 +672,25 @@ describe('client-adapters.test.js', () => {
...
+    ClientAdapters.init()
+
+    let res1 = await ClientAdapters.detectlangGroup.detectlang({
+      method: 'getDetectedLangsList',
+      params: {
+        sourceData: { data: {
+          detections: [
+            { language: 'en', isReliable: true, confidence: 3.36 },
+            { language: 'pt', isReliable: false, confidence: 3.36 },
+            { language: 'eu', isReliable: false, confidence: 3.36 }
```
zfletch, line 686: I'd recommend rewriting this text so that two elements have `isReliable: true`, but with a different `confidence`. That way, it would test that sorting is working correctly.

packages/components/src/lib/controllers/app-controller.js
```diff
@@ -1434,7 +1434,7 @@ export default class AppController {
           return
         }
       }
-      this._store.commit('app/setPrioritizeDefaultLanguage', false)
+      // this._store.commit('app/setPrioritizeDefaultLanguage', false)
```
zfletch, line 1437: Should this be commented out?

packages/client-adapters/tests/adapters/detectlang/adapter.test.js
```diff
@@ -0,0 +1,106 @@
...
+  it('3 DetectLangAdapter - getDetectedLangsList - sourceData - full response', async () => {
+    let adapter = new DetectLangAdapter({
+      category: 'detectlangGroup',
+      adapterName: 'detectlangMethod',
+      method: 'detectLanguage',
+      sourceData: { data: {
+        detections: [
+          { language: 'en', isReliable: true, confidence: 3.36 },
+          { language: 'pt', isReliable: false, confidence: 3.36 },
+          { language: 'eu', isReliable: false, confidence: 3.36 }
```
zfletch, line 56: I'd recommend using a more complicated response where there are multiple `isReliable: false` and multiple `isReliable: true`, the highest confidence has `isReliable: false`, and the other ones have different `confidence`. That would test the language selection logic more fully without requiring  additional tests.

packages/client-adapters/src/adapters/detectlang/adapter.js
```diff
@@ -0,0 +1,84 @@
...
+        const lang = reliableLangs.sort((a, b) => a.confidence.localeCompare(b.confidence, undefined, { numeric: true }))
```
zfletch, line 76: In the test data, `confidence` is a number, not a string. If the API returns a number, there's no reason to use `localeCompare`: you can just use the `<` and `>` functions.

zfletch (changes requested):
This looks pretty good. Most of my comments are just questions or minor suggestions.

I am requesting changes instead of approving because of the two comments in `DetectLangAdapter.chooseOneLanguage`. I think there may be a bug in that function.

packages/client-adapters/package.json
```diff
@@ -8,7 +8,7 @@
     "test": "jest --coverage && shx cat ./coverage/lcov.info | coveralls",
     "test-no-coverage": "jest",
     "test-c": "jest tests/client-adapters.test.js",
-    "test-u": "jest tests/adapters/dtsapi/adapter.test.js",
```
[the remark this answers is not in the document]
irina060981: yes, these npm scripts are used only for developing stage, that's why I change it to the test file that is useful to the current case.

packages/client-adapters/src/adapters/detectlang/adapter.js
```diff
@@ -0,0 +1,84 @@
...
+  constructor (config = {}) {
+    super()
+    this.config = this.uploadConfig(config, DefaultConfig)
+    this.sourceData = config.sourceData
+  }
+
+  /**
+   *
+   * @param {String} text - text for analysis
+   * @returns {String} - langCode ISO-3 - a detected language
```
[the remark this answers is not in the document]
irina060981: I didn't pay much attention to it - thank you for the information - will fix it in comments.

packages/client-adapters/src/adapters/detectlang/adapter.js
irina060981: oh, thank you for the catch

packages/components/src/lib/controllers/app-controller.js
irina060981: No, it is a fix for the language detection for the page - I removed it to fit to updated requirements.
But I decided not to delete it for now - I would delete after testing stage would be finished by Monica.

packages/client-adapters/tests/adapters/detectlang/adapter.test.js
irina060981: thank you for the recomendation - done and method is optimized now :)

packages/client-adapters/src/adapters/detectlang/adapter.js
irina060981: fixed

